PR #3, opened 2016-08-19
listpath --> list

jtratner:
Initially didn't want to call it `list` because it would shadow Python
builtin but it's less confusing to use same name in object API, functional API
and CLI.

Pretty trivial PR, no instances of listpath in tests anyways (I guess since we were testing underlying functional API).

=> @wesleykendall / @phanieste

### DIFF
--- a/stor/__init__.py
+++ b/stor/__init__.py
@@ -61,7 +61,7 @@ expand = _delegate_to_path('expand')
 join = _delegate_to_path('joinpath')
 split = _delegate_to_path('splitpath')
 splitext = _delegate_to_path('splitext')
-listpath = _delegate_to_path('list')
+list = _delegate_to_path('list')
 listdir = _delegate_to_path('listdir')
 glob = _delegate_to_path('glob')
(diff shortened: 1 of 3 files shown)


stor/__init__.py
```diff
@@ -76,6 +76,16 @@ def wrapper(path, *args, **kwargs):
 walkfiles = _delegate_to_path('walkfiles')
 
 
+def list(pth):
+    import warnings
+
+    # DeprecationWarnings are hidden by default. We want to get rid of this
+    # sooner rather than later.
+    warnings.warn('Using the ``list()`` function directly is deprecated,'
+                  ' use ``list()`` instead', UserWarning)
```
wesleykendall: Use ``stor.list()`` instead?

wesleykendall:
@jtratner just had one question. Otherwise lgtm

stor/__init__.py
jtratner: oh I did a bad find/replace crud

jtratner:
wanted to avoid API break

PR #4, opened 2016-08-20
Ignore distribution not found errors

jtratner:
@wesleykendall cc @phanieste - ignores distribution not found errors when not pip installed, making it easier to futz around when working in develop mode.


wesleykendall:
@jtratner lgtm

PR #5, opened 2016-08-20
Clarify top-level glob signature

jtratner:
@wesleykendall cc @phanieste - we accidentally left in toplevel `glob()` in this library.   Since it doesn't really work at the moment (have to pass two arguments that don't really make sense), I just changed the top-level implementation to take in 2 arguments and correctly explain what they mean.

When we implement a single argument glob, we can just allow `pattern=None`.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1,6 +1,11 @@
 Release Notes
 =============
 
+v1.1.1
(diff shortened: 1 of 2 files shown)


wesleykendall:
Thanks @jtratner lgtm

PR #7, opened 2016-10-13
Skip broken symlinks on walk files and dirs

jtratner:
Previously, a broken symlink would be generated by walkfiles, but calling getsize() on it would cause an OSError, borking uploads.  (swift CLI also prints error to console on this, but will still do bulk upload).

This PR handles broken symlink edge case correctly by skipping it in output from walk_files_and_dirs, and returning directories that are empty except for broken symlinks (still rare).

@wesleykendall to review (fixes issue we've been having in production where we were deleting a file during clean up that had a symlink pointing to it, causing upload to fail).

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1,6 +1,12 @@
 Release Notes
 =============
 
+v1.1.2
+------
+
+* Skip broken symlinks during upload, download and listing of files, allowing
+  ``copytree``, ``list`` to work on folders that contain broken symlinks.
+
 v1.1.1
 ------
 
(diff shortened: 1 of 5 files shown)


jtratner:
I also tested this out manually against a broken symlink I created and it worked correctly to upload (just skipping broken symlink altogether)

wesleykendall:
@jtratner lgtm. Looks like walk_files_and_dirs is too complex now if that can be split / simplified in any way

jtratner:
@wes - simplified a conditional that was actually the same and moved the helper function out of the main function (which I guess was counting against the complexity because it shares a namespace?).

PR #9, opened 2016-10-14
[WIP] Ran futurize for fun

kyleabeauchamp:


stor/base.py
```diff
@@ -388,12 +392,12 @@ def listdir(self):
 
         The elements of the list are Path objects.
         """
-        return [self / child
+        return [old_div(self, child)
                 for child in map(self._always_unicode, os.listdir(self))]
 
     def glob(self, pattern):
         cls = self.path_class
-        return [cls(s) for s in glob.glob(self / pattern)]
```
kyleabeauchamp: this one is pretty funny.

kyleabeauchamp: Presumably we should disable that flag and do division changes manually given the abundance of string division.

stor/obs.py
```diff
@@ -276,9 +280,9 @@ def __iter__(self):
     def _buffer(self):
         "Cached buffer of data read from or to be written to Object Storage"
         if self.mode in ('r', 'rb'):
-            return cStringIO.StringIO(self._path.read_object())
+            return io.StringIO(self._path.read_object())
```
lucaswiman: I think these should be `BytesIO`. Object Storage is storing binary blobs of data, not unicode.

kyleabeauchamp: Indeed, good to see failure modes of automatic conversion :)

kyleabeauchamp:
Deprecated by #10

PR #10, opened 2016-10-14
[WIP] Py3k (attempt 2)

kyleabeauchamp:
1.  Added futures to requirements.txt
2.  Ran `futurize  --nofix=division_safe -w stor/*.py stor/*/*.py`
3.  Global find and replace of `io.StringIO` with `io.BytesIO`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ boto3>=1.4.0
 cached_property
 python-keystoneclient>=1.8.1
 python-swiftclient
+futures>=3.0.5; python_version < '3.0'
--- a/stor/base.py
+++ b/stor/base.py
@@ -1,3 +1,5 @@
+from builtins import zip
+from builtins import map
 from stor import utils
 
 import errno
--- a/stor/cli.py
+++ b/stor/cli.py
@@ -72,10 +72,7 @@ subcommand::
 Direct file transfer between OBS services or within one OBS service (server-side copy)
 is not yet supported.
 """
-from future import standard_library
-standard_library.install_aliases()
 from builtins import str
-from builtins import range
 
 import argparse
 import copy
--- a/stor/cli.py
+++ b/stor/cli.py
@@ -72,10 +72,14 @@ subcommand::
 Direct file transfer between OBS services or within one OBS service (server-side copy)
 is not yet supported.
 """
+from future import standard_library
+standard_library.install_aliases()
+from builtins import str
+from builtins import range
 
 import argparse
 import copy
-import ConfigParser
+import configparser
 from functools import partial
 import logging
 import os
@@ -151,7 +155,7 @@ def _get_env():
 
     Returns a ConfigParser.
     """
-    parser = ConfigParser.SafeConfigParser()
+    parser = configparser.SafeConfigParser()
     # if env file doesn't exist, copy over the package default
     if not os.path.exists(ENV_FILE):
         shutil.copyfile(PKG_ENV_FILE, ENV_FILE)
@@ -169,7 +173,7 @@ def _get_pwd(service=None):
     if service:
         try:
             return utils.with_trailing_slash(parser.get('env', service))
-        except ConfigParser.NoOptionError:
+        except configparser.NoOptionError:
             raise ValueError('%s is an invalid service' % service)
     return [utils.with_trailing_slash(value) for name, value in parser.items('env')]
 
@@ -377,7 +381,7 @@ def process_args(args):
         settings.update(settings.parse_config_file(config))
     func_kwargs = {
         key: Path(val) if type(val) is TempPath else val
-        for key, val in args_copy.iteritems() if val
+        for key, val in args_copy.items() if val
     }
     try:
         if pth:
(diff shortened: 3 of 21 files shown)


stor/base.py
```diff
@@ -1,3 +1,5 @@
+from builtins import zip
+from builtins import map
```
lucaswiman: These are pointless, since you're just iterating over these anyway.

requirements.txt
```diff
@@ -2,3 +2,4 @@ boto3>=1.4.0
 cached_property
 python-keystoneclient>=1.8.1
 python-swiftclient
+futures>=3.0.5; python_version < '3.0'
```
lucaswiman: The benefits of futures don't seem be matched by the cost of adding another dependency.

kyleabeauchamp: It's already a dependency: https://github.com/openstack/python-swiftclient/blob/master/requirements.txt

stor/cli.py
```diff
@@ -72,10 +72,14 @@
 Direct file transfer between OBS services or within one OBS service (server-side copy)
 is not yet supported.
 """
+from future import standard_library
+standard_library.install_aliases()
```
lucaswiman: Note that this introduces global monkey-patching, which is a fairly invasive thing for a dependency to do.

requirements.txt
kyleabeauchamp: I'm surprised they use *both* futures and six...

stor/cli.py
```diff
@@ -72,10 +72,14 @@
...
+from builtins import str
+from builtins import range
```
lucaswiman: This is pointless since you're just iterating.

stor/cli.py
```diff
@@ -72,10 +72,14 @@
 Direct file transfer between OBS services or within one OBS service (server-side copy)
 is not yet supported.
 """
+from future import standard_library
+standard_library.install_aliases()
+from builtins import str
```
lucaswiman: Note that most of the calls to `str` in this module are pointless (since they're `'%s' % str(something)`).

jtratner:
Thanks for the PR - I'll take a look next week!

requirements.txt
jtratner: agreed, I think it's possible to do this without futures and rely on test suite to catch it.

stor/cli.py
jtratner: agreed, let's see if we can do without it

stor/cli.py
```diff
@@ -72,10 +72,14 @@
...
+from builtins import range
 
 import argparse
 import copy
-import ConfigParser
+import configparser
```
jtratner: this works? or is this from the aliases thing?

stor/cli.py
jtratner: yeah I'd just get rid of this

stor/obs.py
```diff
@@ -21,7 +25,7 @@ def wrapper(self, *args, **kwargs):
         func = getattr(self._buffer, attr_name)
         return func(*args, **kwargs)
     wrapper.__name__ = attr_name
-    wrapper.__doc__ = getattr(cStringIO.StringIO(), attr_name).__doc__
+    wrapper.__doc__ = getattr(io.BytesIO(), attr_name).__doc__
```
jtratner: potentially better to do:

```
try:
    from cStringIO import StringIO as BytesIO
except ImportError:
    from io import BytesIO
```

just to not change python 2.X behavior...

stor/settings.py
```diff
@@ -1,6 +1,9 @@
+from future import standard_library
+standard_library.install_aliases()
+from builtins import object
```
jtratner: all these lines seem unnecessary

stor/tests/test_integration_s3.py
```diff
@@ -1,4 +1,6 @@
-from cStringIO import StringIO
+from future import standard_library
+standard_library.install_aliases()
+from io import StringIO
```
jtratner: BytesIO?

stor/utils.py
```diff
@@ -1,3 +1,6 @@
+from builtins import next
+from builtins import str
+from builtins import object
```
jtratner: these are prob all unnecessary, no?  or does it change 2.X behavior?

kyleabeauchamp:
FWIW, it appears `_delegate_to_buffer` isn't quite happy with the naive `StringIO` to `BytesIO` conversion due to the lack of `next`.  

```
~/stor/stor/obs.py", line 28, in _delegate_to_buffer
    wrapper.__doc__ = getattr(io.BytesIO(), attr_name).__doc__
AttributeError: '_io.BytesIO' object has no attribute 'next'

```

kyleabeauchamp:
NM, I think I found some comments in the docs about this.

jtratner:
Easiest is to just add an if statement and not define "next" if python 3

kyleabeauchamp:
Hmm, it does look like we have a lot of py2k-style `super(x,y)` calls scattered throughout.  We might want to strategize to deal with those in a uniform way?

kyleabeauchamp:
In particular, our `super` chain eventually seems to lead to calling `object(pth)` on the root object type, which is forbidden in py3k.

jtratner:
I bet it's because python 3 changed behavior when subclassing built in
types (in this case, str)

PR #11, opened 2016-10-17
Retry on UnauthorizedErrors to deal with transient edge cases

jtratner:
@wesleykendall 

Depending on your swift cluster, authentication and authorization can become a bottleneck and you can end up with spurious unauthorized errors in the midst of heavy load.  This PR adds `UnauthorizedError` to list of exceptions to retry on in key methods.

The downside is that, if you set a high number of retries and you _aren't_ authenticated, this change will mean that you get a really slow feedback loop.

What do you think @wesleykendall ? (cc @phanieste if you're interested)

### DIFF
--- a/stor/swift.py
+++ b/stor/swift.py
@@ -180,7 +180,14 @@ class FailedUploadError(UnavailableError):
 
 
 class UnauthorizedError(SwiftError):
-    """Thrown when a 403 response is returned from swift"""
+    """Thrown when a 403 response is returned from swift.
+
+    Note:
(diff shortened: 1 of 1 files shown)


wesleykendall:
Is this documented in swiftstack's docs somewhere? Or something we have experienced? PR lgtm

jtratner:
Something we've experienced - but it may be specific to an auth system being overwhelmed rather than a swift builtin.

Still debating internally whether this is a good idea but #12 helps with debugging these issues when they do occur.

PR #14, opened 2016-10-19
Add six to requirements.txt

kyleabeauchamp:
I didn't add a version because I'm deferring to python-swiftclient to select the version.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1,6 +1,10 @@
 Release Notes
 =============
 
+v1.2.1
+------
+* Add explicit dependence on six to requirements.txt
+
 v1.2.0
 ------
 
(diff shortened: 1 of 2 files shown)


wesleykendall:
@kyleabeauchamp is it erroring out without six? I'm confused why we need it in the requirements if python-swiftclient already has it as a dependency

kyleabeauchamp:
Just for clarity?  We directly call it so IMHO we want it directly included in the requirements.

kyleabeauchamp:
E.g. I didn't even notice that it was being used because it wasn't listed.

wesleykendall:
Cool sounds good to me then. @kyleabeauchamp mind to update the release_notes.rst file under docs? This will go under a 1.2.1 release. Just so there's notes about what we updated.

PR #16, opened 2016-10-20
Switch to new style exception messages

kyleabeauchamp:
Fixes deprecation warning and improves py3k compatibility

### DIFF
--- a/stor/cli.py
+++ b/stor/cli.py
@@ -396,7 +396,7 @@ def process_args(args):
     except exceptions.RemoteError as exc:
         if type(exc) is exceptions.NotFoundError and pth:
             perror('Not Found: %s' % pth)
-        perror('%s: %s\n' % (exc.__class__.__name__, exc.args[0]))
+        perror('%s: %s\n' % (exc.__class__.__name__, str(exc)))
 
 
 def print_results(results):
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
Only question I have is whether we need a `if not None` guard or whether we can assume that we raise the exception with a value (e.g. as in http://stackoverflow.com/questions/13063212/deprecationwarning-baseexception-message-has-been-deprecated-as-of-python-2-6-e)

kyleabeauchamp:
Looking closer, I think the constructor for RemoteError requires having a message present, so there's no need to guard against `None` in this case.

jtratner:
Good catch! :)

Question: why not just use "str(exc)" here? My usual go to is just "%s: %s"
% type(exc).__name__, str(exc)

That to me is the idiomatic way of handling and avoids knowing about
implementation of the exception. (And I believe it covers same purpose here)

kyleabeauchamp:
Done.  Agree that it's more idiomatic, I didn't realize that str/repr already had the correct logic written.

jtratner:
Do you want to merge this?

kyleabeauchamp:
+1 to merge

jtratner:
(didn't realize you couldn't merge it yourself)

PR #19, opened 2017-02-08
Add detection for SwiftStack-s3 style segment containers

jtratner:
## Purpose

SwiftStack's S3 emulation layer creates segment containers that look like `$CONTAINER+segments` on multipart uploads.  This is a third format for segment containers, in addition to swiftclient's `$CONTAINER_segments` and stor's `.segments_CONTAINER`.

I don't see any explicit docs from SwiftStack about it, just happened when I noticed weird `ABCD+segments` containers internally.

## How this was tested

1. Added failing integration test and confirmed it failed
2. Updated code to check for segment container and confirmed integration + unit tests pass
3. Confirmed that `stor list swift://AUTH_sometenant` no longer shows `+segments` containers.

@wesleykendall @kyleabeauchamp @krhaas

### DIFF
--- a/stor/swift.py
+++ b/stor/swift.py
@@ -442,7 +442,9 @@ class SwiftPath(OBSPath):
         """True if this path is a segment container"""
         container = self.container
         if not self.resource and container:
-            return container.startswith('.segments_') or container.endswith('_segments')
+            return (container.startswith('.segments_') or
+                    container.endswith('_segments') or
+                    container.endswith('+segments'))
         else:
             return False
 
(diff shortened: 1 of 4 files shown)


stor/swift.py
```diff
@@ -442,7 +442,9 @@ def is_segment_container(self):
         """True if this path is a segment container"""
         container = self.container
         if not self.resource and container:
-            return container.startswith('.segments_') or container.endswith('_segments')
+            return (container.startswith('.segments_')
+                    or container.endswith('_segments')
```
kyleabeauchamp: Allegedly this or statement is failing our linter.  I'm happy relaxing but up to you.

kyleabeauchamp:
LGTM modulo two comments below:

1.  Linter: https://travis-ci.org/counsyl/stor/jobs/199780255#L541
2.  I can't confirm or deny the need for this extra containing nomenclature, so I defer to you.

stor/swift.py
krhaas: just have to put `or` at end of line above I think.

krhaas:
LGTM too.

PR #21, opened 2017-02-11
Python 3 compatibility

wesleykendall:
@jtratner this builds off of @kyleabeauchamp 's PR (https://github.com/counsyl/stor/pull/10) to provide compatibility to python 3.3 and above. All of my unit tests pass locally. I haven't tried running the integration tests and was wanting to have some help with that.

### Primary changes

1. The weirdness encountered in https://github.com/counsyl/stor/pull/10 wasn't because of inherting from str. It was because the `__init__` signature did not match the `__new__` signature.
2. Most of the complexity in this PR was addressed by returning the appropriate stream depending on the file mode (BytesIO for binary mode and StringIO for non binary)
3. Some of the exceptions were converted with six so that they are compatible with the `raise exception from exception` syntax in Python 3
4. Mocking with autospec on some of the swift calls (like `list`, `upload`, etc) failed because `num_retries` is passed in through the wrapper and is not part of the method signature. I'm not totally sure what changed in how Python 3 obtains signatures, but I went ahead and just used the ` with settings.use(...)` syntax to set `num_retries`.

Everything else should be a pretty straightforward change.

The only thing I'm a little concerned about are possible breakages related to binary vs non-binary modes, but I think the integration test suite will be good enough to catch most of those cases. Might want to add a couple more integration tests for binary and non-binary writes

@jtratner one additional note - the test at https://github.com/counsyl/stor/blob/master/stor/tests/test_s3.py#L1296 failed for me one time in Travis but passed when I re-ran it. I think something might be up in that Threadpool code not propagating the exception properly

### DIFF
--- a/stor/cli.py
+++ b/stor/cli.py
@@ -72,10 +72,7 @@ subcommand::
 Direct file transfer between OBS services or within one OBS service (server-side copy)
 is not yet supported.
 """
-from future import standard_library
-standard_library.install_aliases()
 from builtins import str
-from builtins import range
 
 import argparse
 import copy
--- a/stor/cli.py
+++ b/stor/cli.py
@@ -74,7 +74,6 @@ is not yet supported.
 """
 import argparse
 import copy
-import configparser
 from functools import partial
 import logging
 import os
@@ -84,6 +83,7 @@ import sys
 import tempfile
 
 import six
+from six.moves import configparser
 
 import stor
 from stor import exceptions
--- a/stor/cli.py
+++ b/stor/cli.py
@@ -72,8 +72,6 @@ subcommand::
 Direct file transfer between OBS services or within one OBS service (server-side copy)
 is not yet supported.
 """
-from builtins import str
-
 import argparse
 import copy
 import configparser
@@ -84,6 +82,8 @@ import shutil
 import signal
 import sys
 import tempfile
+
+import six
 
 import stor
 from stor import exceptions
@@ -129,7 +129,7 @@ def _make_stdin_action(func, err_msg):
                 if namespace.func == func:
                     raise argparse.ArgumentError(self, err_msg)
                 else:
-                    ntf = tempfile.NamedTemporaryFile(delete=False)
+                    ntf = tempfile.NamedTemporaryFile(delete=False, mode='w')
                     try:
                         ntf.write(sys.stdin.read())
                     finally:
@@ -170,8 +170,8 @@ def _get_pwd(service=None):
     if service:
         try:
             return utils.with_trailing_slash(parser.get('env', service))
-        except configparser.NoOptionError:
-            raise ValueError('%s is an invalid service' % service)
+        except configparser.NoOptionError as e:
+            six.raise_from(ValueError('%s is an invalid service' % service), e)
     return [utils.with_trailing_slash(value) for name, value in parser.items('env')]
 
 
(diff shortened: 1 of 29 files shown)


stor/obs.py
```diff
@@ -1,7 +1,12 @@
-import cStringIO
 import posixpath
+import sys
 
 from cached_property import cached_property
+from six import BytesIO
+try:
+    from cStringIO import StringIO
+except ImportError:  # pragma: no cover
+    from six import StringIO
```
lucaswiman: IMO it's better to just use the `io.StringIO` class. It's also implemented in C, and has pretty much the same behavior between Python 2 and Python 3. `cStringIO.StringIO` in python 2 cannot handle non-ASCII unicode correctly, and always returns `str`, even when passed unicode:
```python
>>> import io, cStringIO
>>> # ASCII-encodable unicode
... cStringIO.StringIO(u'asdf').getvalue()
'asdf'
>>> io.StringIO(u'asdf').getvalue()
u'asdf'
>>> 
>>> # non-ASCII-encodable unicode
... cStringIO.StringIO(u'💣').getvalue()
  File "<stdin>", line 2, in <module>
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
>>> io.StringIO(u'💣').getvalue()
u'\U0001f4a3'
>>> 
>>> # Bytes
... cStringIO.StringIO('asdf').getvalue()
'asdf'
>>> io.StringIO('asdf').getvalue()
  File "<stdin>", line 1, in <module>
TypeError: initial_value must be unicode or None, not str
```
```

Comment (wesleykendall) [inline stor/obs.py]: Ah I was unaware of its issues handling certain strings. @jtratner any objections to just changing it all to use io.StringIO and io.BytesIO?

Comment (jtratner) [inline stor/obs.py]: sure go for it

**Comment by wesleykendall** [comment]
@jtratner @lucaswiman comments addressed

**Comment by jtratner** [comment]
@wesleykendall - do we have any tests that actually read or write unicode characters? just to make sure we don't hit character encoding issues?

E.g., if I create a UTF-8 encoded document on object storage, does it get correctly read through in Python 2? end up as bytes in Python 3?

**Comment by wesleykendall** [comment]
@jtratner we don't to the best of my knowledge. Can you or anyone else help out on some integration tests for this? I can try some S3 ones later in the week and write a couple unit tests. My main concern is CLI functionality

**Comment by wesleykendall** [comment]
@jtratner can someone there run integration tests? I know quite a few people that want to start using this with python 3

**Comment by jtratner** [comment]
> @jtratner can someone there run integration tests? I know quite a few people that want to start using this with python 3

Integration tests pass in python 2 against swift (side note - shouldn't we update tox to run py3 tests too?)

Did you end up adding a unicode test?

File: stor/cli.py [hunk 2]
```diff
@@ -128,7 +129,7 @@ def __call__(self, parser, namespace, values, option_string=None):
                 if namespace.func == func:
                     raise argparse.ArgumentError(self, err_msg)
                 else:
-                    ntf = tempfile.NamedTemporaryFile(delete=False)
+                    ntf = tempfile.NamedTemporaryFile(delete=False, mode='w')
```
jtratner: so we never exercised this code path before?

jtratner:
@kyleabeauchamp @wesleykendall - got this merge and it should end up on pypi soon

PR #22, opened 2017-03-22
Ensure that OBS file cleans itself up when GC'd

jtratner:
Otherwise you end up not committing files if used (granted, unPythonically) in format `something.to_csv(stor.open(mypath, 'w'))`.

Better to be non-surprising than unclean ;)

I'm totally fine with merging py3 PR before this one if it matters.

@wesleykendall / @krhaas / @kyleabeauchamp to review

cc @pkaleta

### DIFF
--- a/stor/obs.py
+++ b/stor/obs.py
@@ -272,6 +272,9 @@ class OBSFile(object):
         # give users access to underlying buffer)
         return self
 
+    def __del__(self):
+        self.close()
+
     @cached_property
     def _buffer(self):
         "Cached buffer of data read from or to be written to Object Storage"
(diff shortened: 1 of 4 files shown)


stor/obs.py
```diff
@@ -272,6 +272,10 @@ def __iter__(self):
         # give users access to underlying buffer)
         return self
 
+    def __del__(self):
+        if hasattr(self, '_buffer') and not self.closed:
```
krhaas: should the ` if hasattr(self, '_buffer')` condition actually be in `def close()` ?

jtratner: I initially had it in here because I thought it was supposed to be an error to call `close()` twice but I realize that with normal python fileobj you can call close repeatedly so I'll change

krhaas:
LGTM

PR #23, opened 2017-03-30
Support uploading multiple files together with use_manifest=True.

pkaleta:
@jtratner cc @counsyl/stor 

Before this PR, there was no ability to specify multiple files for upload together with `use_manifest=True`. This PR attempts to fix that.

I've added a unit test to make sure that it works okay. Also, I've done multiple hands-on testing of this feature.

### DIFF
--- a/stor/swift.py
+++ b/stor/swift.py
@@ -1132,7 +1132,8 @@ class SwiftPath(OBSPath):
                 Alternatively, when multiple directories are uploaded, manifest file will
                 be created in the current directory. For example::
 
-                    stor.Path('swift://AUTH_foo/bar').upload(['logs', 'test.txt'], use_manifest=True)
+                    stor.Path('swift://AUTH_foo/bar').upload(
+                        ['logs', 'test.txt'], use_manifest=True)
 
                 The manifest will be located at
                 ``swift://AUTH_foo/bar/.data_manifest.csv``
--- a/stor/swift.py
+++ b/stor/swift.py
@@ -1139,8 +1139,6 @@ class SwiftPath(OBSPath):
         """
         if not self.container:
             raise ValueError('must specify container when uploading')
-        if use_manifest and not (len(to_upload) == 1 and os.path.isdir(to_upload[0])):
-            raise ValueError('can only upload one directory with use_manifest=True')
         utils.validate_condition(condition)
 
         swift_upload_objects = [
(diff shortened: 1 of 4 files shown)


jtratner:
Seems reasonable - is there a reason to special case single directory?

Also can you add some examples to test_integration that makes sure it runs the way we expect?  (especially in terms of getting file name correct if we're not uploading to a container and instead to a subdirectory).

Something like:

```
stor.copytree([a, b], 'swift://A/C/some/subpath')
stor.copytree([a, c], 'swift://A/C/some/subpath2')
stor.copytree([a, b, c], 'swift://A/C/some/subpath')

```
where a and b are directories and c is a single file and also maybe determine what we should do if manifest already exists (I'm assuming overwrite?)

pkaleta:
>Seems reasonable - is there a reason to special case single directory?

Yes. The reason to do that is backwards compatibility. The current behavior (`master` branch) is to create the manifest file inside the single directory that was passed. I want to preserve this behavior. So if a user does this:

```
stor.Path('swift://AUTH_foo/bar').upload([`logs`], use_manifest=True)
```

The manifest file will be created at `swift://AUTH_foo/bar/logs`.

In case of uploading multiple files, it's not immediately obvious where to create it. Imagine a situation where a user does this:

```
stor.Path('swift://AUTH_foo/bar').upload(['x.txt', '/tmp', 'logs/2016-02-02.log`], use_manifest=True)
```

There's a couple of ways around this:

- Create manifest in the current directory. This is the solution implemented by this branch. In this case the manifest will be created at `swift://AUTH_foo/bar/.data_manifest.csv`.
- Allow user to specify directory where the manifest is created (e.g. by using `use_manifest=<path to manifest file>`)

I'm going to add a comment that clarifies the current approach.

pkaleta:
>Also can you add some examples to test_integration that makes sure it runs the way we expect? 

Good call. I've added integration test case in 8ad476ebef1933b1434361261355bd1a8ad1fc25.

pkaleta:
friendly ping @jtratner

stor/swift.py
```diff
@@ -1155,8 +1169,15 @@ def upload(self,
         # Convert everything to swift upload objects and prepend the relative
         # resource directory to uploaded results. Ignore the manifest file in the case of
         # since it will be uploaded individually
-        manifest_file_name = (Path(to_upload[0]) / utils.DATA_MANIFEST_FILE_NAME
-                              if use_manifest else None)
+        if use_manifest:
+            if len(to_upload) == 1 and os.path.isdir(to_upload[0]):
+                manifest_path_prefix = Path(to_upload[0])
+            else:
+                manifest_path_prefix = Path('.')
```
jtratner: and this is because we change directory before doing the uploading?

pkaleta:
>Aside from the location of the manifest, these two commands would/do end up with the same files uploaded, right?

Correct.

>just in the second data manifest goes to swift://AUTH_foo/bar/.data_manifest.csv)?

Yes, however you'd have to pass `use_manifest=True` to `.upload()` for the manifest to be created.

stor/swift.py
pkaleta: Correct.

pkaleta:
Thanks @jtratner

pkaleta:
@jtratner will merge, because I don't have permissions.

PR #27, opened 2017-05-17
Force tox to actually run 3.5 tests

### LINKED ISSUE #26: Actually run python 3.5 tests with tox
Because coverage is whitelisted, it's actually just running python 2.7 tests with the 3.5 interpreter. womp womp :P

jtratner:
This only matters for internal Counsyl CI.

And remove python 3.4 which I don't think we have installed internally.
(since Travis runs it, not a big deal)

Fixes #26 

@kyleabeauchamp @wesleykendall @pkaleta  to review

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 install_command = pip install {opts} {packages}
 indexserver =
     default = https://pypi.python.org/simple/
-envlist = py27,py34,py35
+envlist = py27,py35
 
 [testenv]
 usedevelop = True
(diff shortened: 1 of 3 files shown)


kyleabeauchamp:
LGTM

tox.ini
```diff
@@ -15,7 +15,6 @@ deps =
   -r{toxinidir}/requirements-dev.txt
 whitelist_externals = make
                       bash
-                      coverage
                       nosetests
```
lucaswiman: I think the whole `whitelist_externals` section is unneeded. It's amazing how many places this code has propagated 😩

PR #29, opened 2017-05-17
Add stor.utils.is_writeable.

pkaleta:
@jtratner 

The goal of `stor.utils.is_writeable` is to be able to determine whether we have write access to the specified path, either local, S3 or Swift. This is useful, because using this we'd be able to check for write permissions ahead-of-time, before generating objects for upload, which might take significant amount of time.

Check the docstring and tests for more details.

### DIFF
--- a/stor/utils.py
+++ b/stor/utils.py
@@ -268,6 +268,7 @@ def is_writeable(path, swift_retry_options=None):
     from stor import rmtree
     from stor.swift import SwiftPath
     from stor.swift import UnauthorizedError
+    from stor.swift import UnavailableError
 
     path = with_trailing_slash(Path(path))
 
@@ -294,7 +295,7 @@ def is_writeable(path, swift_retry_options=None):
             # Attempt to create a file in the `path`.
             copy(tmpfile.name, path, swift_retry_options=swift_retry_options)
             answer = True
-        except (UnauthorizedError, IOError, OSError, exceptions.FailedUploadError):
+        except (UnauthorizedError, UnavailableError, IOError, OSError, exceptions.FailedUploadError):  # nopep8
             answer = False
         else:
             # Remove the file that was created.
--- a/stor/utils.py
+++ b/stor/utils.py
@@ -265,7 +265,7 @@ def is_writeable(path, swift_retry_options=None):
     from stor import join
     from stor import Path
     from stor import remove
-    from stor import rmtree
+    from stor.swift import ConflictError
     from stor.swift import SwiftPath
     from stor.swift import UnauthorizedError
     from stor.swift import UnavailableError
@@ -303,8 +303,11 @@ def is_writeable(path, swift_retry_options=None):
     # Remove the Swift container if it didn't exist when calling this function, but exists
     # now. This way this function remains a no-op with regards to container structure.
     if container_existed is False and container_path.exists():
-        assert not container_path.listdir()
-        rmtree(container_path)
+        try:
+            container_path.remove_container()
+        except ConflictError:
+            # Ignore if some other thread/user created the container in the meantime.
+            pass
 
     return answer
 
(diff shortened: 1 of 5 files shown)


jtratner:
@pkaleta - can you rebase this on top of current master? It fixes up tests so that `make test` locally actually tests python 3 code too.

pkaleta:
>@pkaleta - can you rebase this on top of current master? It fixes up tests so that make test locally actually tests python 3 code too.

Done. Tests passed locally.

stor/utils.py
```diff
@@ -229,6 +230,85 @@ def is_obs_path(p):
...
+    from stor.swift import UnauthorizedError
+
+    path = Path(path)
+
+    # We want this function to behave as a no-op with regards to the underlying directory
+    # structure. Therefore we need to remove any directories created by this function that
+    # were not present when it was called. The `base_path_existed` defined below will
+    # store whether the directory that we're checking existed when calling this function,
+    # so that we know if it should be removed at the end.
+    if is_filesystem_path(path):
```
pkaleta: >could we just use os.access for is_filesystem_path? That way we don't have to deal with the following situation:

I think it's the simplest approach. I could calculate what's the inner-most directory that exists before calling the function, but I think it makes sense for it to behave the same way as `osa

>This means additionally that the following doesn't work for filesystems:

`is_writeable(p)` returning true wasn't supposed to mean that you can immediately write files to `p`, but rather that you're allowed to create the directory structure described by `p` and then store files there. I think it makes sense to keep the behavior consistent with `os.access` in order not to confuse users who're used to `os.access` behavior.

Fixed in 773bafaa30d542f1df558ee7616b73f724b030ec

stor/utils.py
```diff
@@ -229,6 +230,85 @@ def is_obs_path(p):
...
+
+    with tempfile.NamedTemporaryFile() as tmpfile:
+        try:
+            if is_filesystem_path(path):
+                make_dest_dir(path)
+
+            # Attempt to create a file in the `path`.
+            copy(tmpfile.name, path)
+            answer = True
+        except (UnauthorizedError, IOError, OSError, exceptions.FailedUploadError):
```
pkaleta: `copy` calls `SwiftPath.upload`, which already is decorated with `_swift_retry` decorator that implements exponential backoff.

>though we might want to consider allowing user to say "I want a fast is_writeable here!"

bff3dd6c97033fe8ed4ca38b9aca5a4d72f66ea8

stor/utils.py
```diff
@@ -229,6 +230,85 @@ def is_obs_path(p):
...
+            answer = False
+        else:
+            # Remove the file that was created.
+            remove(join(path, basename(tmpfile.name)))
+
+    # Remove the base directory if it didn't exist when calling this function, but exists
+    # now. This way the underlying directories should remain untouched.
+    if base_path_existed is False and base_path.exists():
+        assert not base_path.listdir()
+        rmtree(base_path)
```
pkaleta: >for swift we should consider issuing a direct swift-API container delete (i.e., NOT rmtree, which will list the container and then delete, but instead make a swift service call like here - https://github.com/counsyl/stor/blob/e535729d85aacbf3b223d83e0daf44a86096ead6/stor/swift.py#L1309 )

rmtree doesn't list contents if you pass `swift://tenant/container`. It only does that if you pass the directory as well: `swift://tenant/container/dir` ([check out this if](https://github.com/counsyl/stor/blob/e535729d85aacbf3b223d83e0daf44a86096ead6/stor/swift.py#L1308)). Therefore this is effectively already doing what you're asking for, I think.

>this still has some race conditions - not sure if it's really worth doing this (maybe we could add a OBS-only "rmdir" functionality that only removes empty containers and directory markers?)

Agree, I'll add that.

pkaleta: @jtratner done. WDYT?

pkaleta:
Ping @jtratner. I've fixed addressed all your previous comments.

jtratner:
Cool - lgtm :)

pkaleta:
Thanks @jtratner

krhaas:
👍

PR #30, opened 2017-06-28
Fix SwiftPath.dirname() for top-level swift paths.

pkaleta:
@jtratner 

There's a bug that occurs when calling `dirname()` on top-level swift paths like `swift://` or `swift://a`:

```
  File "/Users/piotr/counsyl/stor/stor/tests/test_swift.py", line 77, in test_dirname_top_level
    self.assertEquals(p1.dirname(), 'swift://')
  File "/Users/piotr/counsyl/stor/stor/base.py", line 151, in dirname
    return self.path_class(self.path_module.dirname(self))
  File "/Users/piotr/counsyl/stor/stor/obs.py", line 71, in __init__
    raise ValueError('path must have %s (got %r)' % (self.drive, pth))
ValueError: path must have swift:// (got u'swift:')
```

I've fixed it by making stor return `swift://` in those cases. The alternative would be to return an empty string or `None`, but I think the current approach is more consistent with the current behavior of `os` module, where it returns `C:\` and `/` for windows and linux top-level dirs respectively.

I've added a test that demonstrates the bug and made sure that it's fixed with this patch in place.

### DIFF
--- a/stor/obs.py
+++ b/stor/obs.py
@@ -1,3 +1,4 @@
+import os
 import posixpath
 import sys
 
@@ -73,6 +74,14 @@ class OBSPath(Path):
 
     copy = utils.copy
     copytree = utils.copytree
+
+    def dirname(self):
+        """
+        Return the directory name of self.
+        """
+        assert self.startswith(self.drive)
+        rest = self[len(self.drive):]
+        return type(self)(self.drive + os.path.dirname(rest))
 
     def __repr__(self):
         return '%s("%s")' % (type(self).__name__, self)
--- a/stor/obs.py
+++ b/stor/obs.py
@@ -81,7 +81,7 @@ class OBSPath(Path):
         """
         assert self.startswith(self.drive)
         rest = self[len(self.drive):]
-        return self.path_class(self.drive + os.path.dirname(rest))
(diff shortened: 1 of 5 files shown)


pkaleta:
A similar bug was present in S3's `dirname` so I fixed it here too.

stor/obs.py
```diff
@@ -74,6 +75,14 @@ def __init__(self, pth):
     copy = utils.copy
     copytree = utils.copytree
 
+    def dirname(self):
+        """
+        Return the directory name of self.
+        """
+        assert self.startswith(self.drive)
+        rest = self[len(self.drive):]
+        return type(self)(self.drive + os.path.dirname(rest))
```
jtratner: I *think* this needs to be self.path_class or path_module? not totally clear

jtratner: (because on windows you still want to use posixpath dirname semantics)

pkaleta: Done. Previous version was working as well, but I think this one is more explicit.

jtratner:
lgtm once tests pass and add note to release notes if you haven't already

jtratner:
(use squash & merge when you merge it ;) )

PR #34, opened 2017-07-07
Enable multiple SOURCE(s) for ``stor cp``

kristjaneerik:
New usage: ``stor cp [-h] [-r] SOURCE [SOURCE ...] DEST``

Also adds ``stor cpto`` which reverses DEST and SOURCE(s) and is useful for e.g. `find . -name '*.csv' | xargs stor cpto swift://tenant/container/prefix/`.

Enable Swift integration to be more flexible by adding more environment variables to control e.g. the tenant.

Note: Somewhat in progress -- `stor.tests.test_integration_swift.SwiftIntegrationTest.test_copytree_multiple_to_from_dir` might not work correctly and `stor cpto` doesn't have tests at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,9 @@ positional arguments:
 
     list                List contents using the path as a prefix.
     ls                  List path as a directory.
-    cp                  Alias for copy.
+    cp                  Copy a source to a destination path.
+    cpto                Copy a source to a destination path. Note that here
+                        DEST comes before SOURCE(s).
     rm                  Remove file at a path.
     walkfiles           List all files under a path that match an optional
                         pattern.
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -8,7 +8,7 @@ To run all tests, type::
 
     make test
 
-In order to run swift integration tests, create a swift tenant called ``AUTH_swft_test`` and provide environment variables for a user that has permissions to write to that tenant (``SWIFT_TEST_USERNAME`` and ``SWIFT_TEST_PASSWORD``). Also set the swift auth url environment variable (``OS_AUTH_URL``)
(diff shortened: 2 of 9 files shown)


jtratner:
What do you think about aping the current man page for copy and adding a `'-t', '--target-directory'` flag and overriding the default help to look like this instead (so again it's more similar to POSIX copy):

```
usage: stor cp [-h] [-r] SOURCE DEST
            stor cp [-r] -t DEST SOURCE

Alias for copy. '-' is a special character that allows for using stdin as the
source.

optional arguments:
  -h, --help  show this help message and exit
  -r          Copy a directory and its subtree to the destination directory.
              Must be specified before any other flags.
```

jtratner:
how do we want to handle mix of directories and files?

kristjaneerik:
It _would_ be really neat to have the `-t` argument, but I tried to express it in `argparse` and couldn't do it -- too hard to not have the positional DEST if the `-t` argument is set, at least with my (limited) knowledge of `argparse`. The way I did it would give you something like `stor cp [-r] [-t DEST] SOURCE [SOURCE ...] [DEST]` and you'd have to do validation of args after getting them, rather than having `argparse` do it.

I would like for it to handle both files and directories (just call `isdir()` on each `SOURCE` and choose appropriately? Is that expensive, does it matter?), but right now it has to be all one or the other, good catch. That way we could remove the `-r` flag, right?

kristjaneerik:
I changed this to be a general CLI enhancement PR since many of the changes are in the same vain.

WIP!

jtratner:
if you output something prefixed as `file:///` does that work with stor? Just looking at your example...

jtratner:
Just thinking about this some more: let's see if we can get this to look really nice and handle stat output well for s3 and swift, and then I can work independently on getting this to work in a performant manner with many objects.

jtratner:
One thing - the openstack object storage API (aka swift) actually returns a list of JSON objects that includes date modified and size, so you don't have to stat each file individually. We just didn't care about it in v1, so we throw away all that information. My suggestion would be to use that info - will be *much* *much* faster on the front end vs calling getsize() repeatedly.

I'm thinking we could add a caching contextmanager that only maintains the cache for the duration of the operation, so user never sees stale results) - especially would matter for not accidentally caching on retry.  Or just stuff the data on to each object's dictionary when we return from list...

But it's probably easier to get this into a good state in terms of parameters and formatting (aiming to have defaults be what comes back from list command by default last-modified, size, (storage class?) for S3 and last-modified, size, content-type for Swift), then to harp on performance enhancement (I can try to do that later)

PR #40, opened 2017-07-10
Fix segment container listing to ignore anything that starts with .segments

jtratner:
relaxing it just means we truly won't break anything

cc @pkaleta @krhaas @wesleykendall @kyleabeauchamp


krhaas:
LGTM

pkaleta:
LGTM.

PR #45, opened 2017-09-27
Fix py3k, add test for opening in various filemodes

kyleabeauchamp:
1.  Add tests opening stor objects in different modes and writing different byte and strings
2.  Fixes issues with py3k and OBSFile

The new tests currently require `TEST_SWIFTPATH` be set via the environment, but we can possible recycle the existing integration test framework for this.  Regardless, they default to not running when the env is not set up.


kyleabeauchamp:
Also: currently this only works on py3k, but presumably that's the missing piece right now so we can adjust as needed to meet in the middle.

kyleabeauchamp:
OK, I've made a number of additional changes to the OBS class and the integration tests so that the tests work on py27 and py36.  On my machine, the only remaining failures seem to be unrelated (see log below).  One remaining question is exactly what we want to do (harden vs discard) with my new tests.

```
======================================================================
ERROR: test_temp_url (stor.tests.test_integration_swift.SwiftIntegrationTest)
----------------------------------------------------------------------
  File "~/stor/stor/tests/test_integration_swift.py", line 148, in test_temp_url
    basic_temp_url = basic_obj.temp_url(inline=False, filename=basic_file)
  File "~/stor/stor/swift.py", line 633, in temp_url
    'a temporary url key must be set with settings.update '
ValueError: a temporary url key must be set with settings.update or by setting the OS_TEMP_URL_KEY environment variable
-------------------- >> begin captured logging << --------------------
stor.swift.progress: INFO: starting upload of 2 objects
stor.swift.progress: INFO: upload complete - 2/2	0:00:00	0.00 MB	0.00 MB/s
--------------------- >> end captured logging << ---------------------

======================================================================
FAIL: test_push_metadata (stor.tests.test_integration_swift.SwiftIntegrationTest)
----------------------------------------------------------------------
  File "~/stor/stor/tests/test_integration_swift.py", line 318, in test_push_metadata
    self.assertEqual(stat_data['Read-ACL'], '.r:*')
AssertionError: '' != '.r:*'
-------------------- >> begin captured logging << --------------------
stor.swift.progress: INFO: starting upload of 1 objects
stor.swift.progress: INFO: upload complete - 1/1	0:00:10	0.00 MB	0.00 MB/s
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
```

kyleabeauchamp:
Tests passing!

stor/tests/test_integration.py
```diff
@@ -44,6 +47,12 @@ def assertCorrectObjectContents(self, test_obj_path, which_test_obj, min_obj_siz
                 expected = self.get_dataset_obj_contents(which_test_obj, min_obj_size)
                 self.assertEquals(contents, expected)
 
+        def _skip_if_filesystem_python3(self, dirname, explanation='needs bytes/str fix'):
```
jtratner: so you should then remove this and the check 'cuz your PR fixes it :)

stor/tests/test_integration_filemodes.py
```diff
@@ -0,0 +1,81 @@
...
+        fp = self.open(self.test_filename, mode='wb')
+        fp.write(byte_string)
+
+    @skipIf(not six.PY3, "Only tested on py3")
+    @raises(TypeError)
+    def test_write_string_to_binary(self):   # pragma: no cover
+        fp = self.open(self.test_filename, mode='wb')
+        fp.write(string_string)
+
+    @skipIf(not six.PY3, "Only tested on py3")
```
jtratner: you can just add the appropriate ones on to the base integration test (probably only need one or two of them) and then I'd suggest making it:

```
bytes_string = six.bytes_type(b'string')
string_string = six.text_type('string')
```

and then finishing the write.  I'd also like it if you could include an encoded unicode character in the bytes string and a decoded unicode character in the unicode string, just to uncover concatenation errors.

stor/obs.py
```diff
@@ -293,7 +293,10 @@ def stream_cls(self):
     def _buffer(self):
         "Cached buffer of data read from or to be written to Object Storage"
         if self.mode in ('r', 'rb'):
-            return self.stream_cls(self._path.read_object())
+            try:  # Catch possible bytes / str issues on py3k
+                return self.stream_cls(self._path.read_object())
```
jtratner: ah okay, so I was expecting that read_object always returns bytes in Python 3, so then we decode to utf-8 if you open it with 'w' or 'r' - right? I think that's how the filesystem works in Python 3?

kyleabeauchamp:
There is only one failure, which seems like a possible sporadic network failure?  See log below:

```
======================================================================
FAIL: test_download_other_error (stor.tests.test_s3.TestDownload)
----------------------------------------------------------------------
  File "/home/travis/build/counsyl/stor/.venv/local/lib/python2.7/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/home/travis/build/counsyl/stor/stor/tests/test_s3.py", line 1312, in test_download_other_error
    S3Path('s3://bucket/path').download('test')
AssertionError: ValueError not raised
-------------------- >> begin captured logging << --------------------
stor.s3.progress: INFO: starting download of 3 objects
stor.s3.progress: INFO: download complete - 1/3	0:00:00	1.00 MB	1.00 MB/s
--------------------- >> end captured logging << ---------------------
```

kyleabeauchamp:
I re-trigged build, which is now green!11

jtratner:
Thanks for this PR @kyleabeauchamp !  This is a really great start towards full python 3 compatibility!

PR #49, opened 2017-10-02
Run posix in integration tests

jtratner:
@pkaleta @kyleabeauchamp - sets up the stor integration tests to run against POSIX as well.  This would have surfaced the issues with `stor.open()` in Python3 land way earlier :)

For now, I've added a skip test to the tests that fail in Python 3, so then @kyleabeauchamp can remove them after his PR merges and fixes the underlying problem.

(In my estimation @kyleabeauchamp I think you'd just need to add one test that opens with 'wb' and another that opens just with 'w')

I'd like to get these tests to pass (with whatever skips) as a minimal PR and then use this in your work @kyleabeauchamp .

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1,5 +1,10 @@
 Release Notes
 =============
+Latest
+------
+* Auto-create parent directories on the filesystem for `stor.copy`, `stor.open`, and `stor.copytree`.
+* Allow `stor.copytree` to work if it targets an empty target directory (removes the other directory first)
+
 
 v1.4.5
 ------
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -2,8 +2,8 @@ Release Notes
 =============
 Latest
 ------
-* Auto-create parent directories on the filesystem for `stor.copy`, `stor.open`, and `stor.copytree`.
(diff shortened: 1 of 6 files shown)


jtratner:
Err - rephrase: I want these to pass in Python 2, we can merge it and then fix the broken build with Kyle's branch after the fact.

Seems like the integration tests were quite broken against Python 3... :(

I've also tweaked the regression tests to be less counsyl-specific for swift: they now let you specify a TEST_TENANT too.

I've paused the deploy pipeline for stor so we can get all the changes together to have it *really* work with Python 3 and then cut a new release then.

kyleabeauchamp:
Which is the change that enables posix?

jtratner:
forgot to commit the file - my bad

kyleabeauchamp:
LGTM % 🍏

kyleabeauchamp:
Lint failures:

```
stor/tests/test_integration_posix.py:5:1: E302 expected 2 blank lines, found 1
stor/tests/test_integration_posix.py:27:1: W391 blank line at end of file
stor/tests/test_integration.py:14:1: E303 too many blank lines (3)
stor/tests/test_integration.py:52:9: E303 too many blank lines (2)
make: *** [lint] Error 1
```

PR #53, opened 2017-11-09
Fix Posix + S3 integration tests

jtratner:
@kyleabeauchamp @pkaleta - these changes will allow us to deploy #45 finally.

Changes:

1. Allow overriding settings for S3, so we can explicitly change s3 integration tests to pull from TEST environment variables. (+ add some tests for it)
2. Change `posix.walkfiles()` to check for isfile() so that it's consistent with all other `walkfiles()` methods
3. Allow S3 Integration tests to pass by fixing regex test

### DIFF
--- a/stor/tests/test_integration_s3.py
+++ b/stor/tests/test_integration_s3.py
@@ -20,24 +20,29 @@ class S3IntegrationTest(BaseIntegrationTest.BaseTestCases):
     bucket on S3.
 
     In order to run the tests, you must have valid AWS S3 credentials set in the
-    following environment variables: AWS_DEFAULT_REGION, AWS_ACCESS_KEY_ID,
-    AWS_SECRET_ACCESS_KEY. In order to have the tests actually run, you also need
-    to additionally set the AWS_TEST_ACCESS_KEY_ID environment variable. For now,
-    it can simply be set to any value.
+    following environment variables: AWS_TEST_ACCESS_KEY_ID,
+    AWS_TEST_SECRET_ACCESS_KEY (and optionally AWS_DEFUALT_REGION).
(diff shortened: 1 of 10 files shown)


stor/tests/test_integration_s3.py
```diff
@@ -20,24 +20,29 @@ class S3IntegrationTest(BaseIntegrationTest.BaseTestCases):
     bucket on S3.
 
     In order to run the tests, you must have valid AWS S3 credentials set in the
-    following environment variables: AWS_DEFAULT_REGION, AWS_ACCESS_KEY_ID,
-    AWS_SECRET_ACCESS_KEY. In order to have the tests actually run, you also need
-    to additionally set the AWS_TEST_ACCESS_KEY_ID environment variable. For now,
-    it can simply be set to any value.
+    following environment variables: AWS_TEST_ACCESS_KEY_ID,
+    AWS_TEST_SECRET_ACCESS_KEY (and optionally AWS_DEFUALT_REGION).
```
kyleabeauchamp: typo.  Normally I'd let it slide but it's an env variable name so I must be a pedant.

kyleabeauchamp:
LGTM modulo typo fix

jtratner:
I was testing this and realize it actually introduces a regression on python 2 where it would silently allow other encodings

```
>>> text = u'😸'
>>> with open('myfile.txt', 'w') as fp:
...    fp.write
  File "<stdin>", line 2, in <module>
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
```

will have a PR soon that makes sure it emulates this behavior in python 2

PR #54, opened 2017-11-09
Ensure we respect locales when reading/writing files

jtratner:
@kyleabeauchamp @pkaleta - the implementation that I (accidentally) allowed to be deployed to PyPI as v1.5.0 forces all text to be utf-8 on OBS, which probably isn't the behavior we want to show (works on Python 3, but only if utf-8 is the encoding).

I still need to fix a couple failing tests, but I wanted to get feedback on whether this makes sense and/or is the right way to do it.

This PR:

1. adds an `encoding` argument (only for Python3 ) that allows explicitly setting encoding
2. falls back to `locale.getpreferredencoding(False)` if encoding not set
3. Confirms that `write_object()` and `read_object()` should handle bytes only. (and adds a deprecation warning for the future)

### DIFF
--- a/stor/cli.py
+++ b/stor/cli.py
@@ -168,10 +168,9 @@ def _get_pwd(service=None):
     or all services if none specified.
     """
     def to_text(data):
-        if six.PY2:
-            return data.decode(locale.getpreferredencoding(False))
-        else:
-            return data
+        if six.PY2:  # pragma: no cover
+            data = data.decode(locale.getpreferredencoding(False))
+        return data
 
     parser = _get_env()
     if service:
--- a/stor/cli.py
+++ b/stor/cli.py
@@ -75,6 +75,7 @@ is not yet supported.
 import argparse
 import copy
 from functools import partial
+import locale
 import logging
 import os
 import shutil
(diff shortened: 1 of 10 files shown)


jtratner:
(I removed the 1.5.0 release that should not have gone out because of buggy locale behavior - we will release this as 1.6.0 just to avoid confusion)

stor/cli.py
```diff
@@ -166,13 +167,19 @@ def _get_pwd(service=None):
...
+
     parser = _get_env()
     if service:
         try:
-            return utils.with_trailing_slash(parser.get('env', service))
+            return to_text(utils.with_trailing_slash(parser.get('env', service)))
         except configparser.NoOptionError as e:
             six.raise_from(ValueError('%s is an invalid service' % service), e)
-    return [utils.with_trailing_slash(value) for name, value in parser.items('env')]
+    return [to_text(utils.with_trailing_slash(value)) for name, value in parser.items('env')]
```
kyleabeauchamp: Don't fully understand this change yet.  Is there a bug report or test regarding this?

jtratner: ConfigParser generates bytes, but we want text

jtratner: it hits the assertionerror later on if you don't do this

kyleabeauchamp:
LGTM.  Presumably we want thorough integration testing if we're adding to *args.

PR #55, opened 2017-12-05
Fix up python 3-compatible stor for release

jtratner:
### Changes

1. Fix up test case mocks to use bytes for body data when they ought to do that
2. Change virtualenv generation so it actually respects the version that travis has sourced. (e.g., on prior builds it'd list out `python 3.6` as the global python, but then virtualenv would go pick `python2` by default when generating a venv).
3. Remove Python 3.3 Travis test because sphinx no longer supports it.

### Background

The (now-removed-from-PyPI) stor v1.5.0 had a bug in it as described in #54.  In fixing that bug, we changed to a more explicit way of handling read/write buffers (i.e., actually using type information from read/write) and using correct locale-encoding for OBS.  This introduced some additional errors, which were caught in the internal deploy pipeline and so that version was never released.

### How were the bugs missed?

Turns out Travis was not running Python 3 tests!  Also I assumed `make test` would pass locally if Travis was passing.

### Steps to prevent this in the future

- [x] ensure Travis runs python 3 tests
- [ ] make it easier to run arbitrary revisions with integration test (will leave for follow up PR)

### How was this tested

* ran `make test` without credentials set
* ran `make test` with test credentials for S3 and Swift set
* Travis tests passed
* manually played around in the shell briefly

Opening PR early so we can see if it'll pass on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 language: python
 python:
   - '2.7'
-  - '3.3'
   - '3.4'
   - '3.5'
   - '3.6'
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 PACKAGE_NAME=stor
 TEST_OUTPUT?=nosetests.xml
 PIP_INDEX_URL=https://pypi.python.org/simple/
+PYTHON?=$(shell which python)
 
 ifdef TOX_ENV
 	TOX_ENV_FLAG := -e $(TOX_ENV)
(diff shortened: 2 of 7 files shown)


jtratner:
@kyleabeauchamp @pkaleta -  this is ready for you to review

Here's updated issue description:

### Changes

1. Fix up test case mocks to use bytes for body data when they ought to do that
2. Change virtualenv generation so it actually respects the version that travis has sourced. (e.g., on prior builds it'd list out `python 3.6` as the global python, but then virtualenv would go pick `python2` by default when generating a venv).
3. Remove Python 3.3 Travis test because sphinx no longer supports it.

### Background

The (now-removed-from-PyPI) stor v1.5.0 had a bug in it as described in #54.  In fixing that bug, we changed to a more explicit way of handling read/write buffers (i.e., actually using type information from read/write) and using correct locale-encoding for OBS.  This introduced some additional errors, which were caught in the internal deploy pipeline and so that version was never released.

### How were the bugs missed?

Turns out Travis was not running Python 3 tests!  Also I assumed `make test` would pass locally if Travis was passing.

### Steps to prevent this in the future

- [x] ensure Travis runs python 3 tests
- [ ] make it easier to run arbitrary revisions with integration test (will leave for follow up PR)

### How was this tested

* ran `make test` without credentials set
* ran `make test` with test credentials for S3 and Swift set
* Travis tests passed
* manually played around in the shell briefly

Opening PR early so we can see if it'll pass on Travis.

kyleabeauchamp:
LGTM

PR #56, opened 2017-12-10
Set universal flag now that we are python2/3 compatible

jtratner:
Right now we only build `py2-none-any.whl`.  Now that we are python 3 compatible (for real!), we should put up `py2.py3-none-any.whl` so Python 3 can benefit from wheels too!


kyleabeauchamp:
According to https://wheel.readthedocs.io/en/stable/, this LGTM

PR #57, opened 2018-01-09
DRY Exception Hierarchy

jtratner:
When we moved exceptions into a single location, we failed to move *all*
of the overlapping swift exceptions out. Before this PR, you'd get
strange behavior like this:

```
try:
   stor.copy(src, dest)
except stor.exceptions.UnavailableError:
   pass
except stor.swift.UnavailableError:
   # oh noes!
```

This PR fixes up the hierarchy a little bit to make it all a little more
cohesive. It does not cause any API breaks.

cc @wesleykendall @pkaleta @kyleabeauchamp

### DIFF
--- a/stor/exceptions.py
+++ b/stor/exceptions.py
@@ -25,7 +25,15 @@ class NotFoundError(RemoteError):
 
 
 class UnauthorizedError(RemoteError):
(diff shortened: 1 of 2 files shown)


wesleykendall:
Nice, lgtm!

PR #58, opened 2018-01-19
Expose stor.utils.is_obs_path

jtratner:
Really simple - makes `is_obs_path` part of top-level API of stor (this was an omission from earlier).

@wesleykendall or @kyleabeauchamp if you have a sec

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1,6 +1,11 @@
 Release Notes
 =============
 
+v1.5.3
+------
+
+* Hoist ``stor.utils.is_obs_path`` --> ``stor.is_obs_path``
+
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
The API change SGTM, but what's the deal with additional guard+ValueError?  It seems like it's checking both source and dest, but printing an exception only about destionation.

jtratner:
good catch @kyleabeauchamp - my bad was for another PR 😅

jtratner:
now only changes `__init__.py` and release notes

kyleabeauchamp:
LGTM

jtratner:
Thanks @kyleabeauchamp !

PR #59, opened 2018-01-19
Refactor tests and add documentation

jtratner:
Pre-cursor to being able to support OBS -> OBS copying.

All of this is essentially a no-op change.  Follow up PR (that I'll open in a second) shows need for it (basically I needed to do a bunch of combinatorial tests with different sources and targets and it made it easier to refactor a bit to do it, rather than rewrite all the tests multiple times)

cc @pkaleta @kyleabeauchamp @wesleykendall


jtratner:
All the tests pass locally for me - just have to fix coverage.

wesleykendall:
@jtratner there isn't really any new code here right, just refactoring? It looks good to me. I'm excited to hear about OBS -> OBS copying coming!

jtratner:
Yep - just makes this diff smaller and allows for the extra Swift -> S3, S3 -> S3, Swift -> Swift tests. 
https://github.com/jtratner/stor/compare/refactor-tests...jtratner:copy-obs-to-obs

and then in a follow up PR we can implement S3 -> S3 server side copy

PR #60, opened 2018-01-19
Add makedirs_p to cross-compatible API

jtratner:
(small PR) essentially just a no-op for OBS paths, but it's super annoying not to have it available.

I didn't add `mkdir`, `mkdir_p`, or `makedirs` because they all raise errors if paths exists or intermediate paths don't exist and it seems really pointless to re-implement all that functionality for no reason.

=> @kyleabeauchamp or @kristjaneerik if either of you folks have time for a quick review.

This is (sort of) another step for enabling OBS -> OBS copying


kristjaneerik:
LGTM

PR #62, opened 2018-01-20
 More aggressively validate ambiguous file paths

### LINKED ISSUE #52: Don't allow writing target name to end with trailing slash
I accidentally forgot to define the filename at the end of a path and did something like:
```python
with stor.open('swift://A/C/path/', 'w') as fp:
    pass
```
with there already being objects with the prefix `swift://A/C/path/`.
This caused the HTTPS access page to not render, it served me the file instead.
`stor.remove('swift://A/C/path/')` removed the object and all was good.

@jtratner suggests that this should just not be allowed happen -- if a copy or write target ends with a trailing slash, the command should fail.

jtratner:
Closes #52

* Error on filepaths with trailing slash because they do not have a name
  and thus appear to be directories.
* Add validation for bad paths to ``OBSFile``.
* Warn if filesystem paths would not pass this validation (otherwise you
  have the possibility of something that works locally and not on OBS).
  This can be set to raise instead if you set
  ``stor.raise_on_ambiguous_filesystem_path`` to be True
* Deprecate ``is_ambiguous()`` which didn't 100% serve the purpose we
  wanted anyways (since you can always tell when user means directory
  because of the user of ``copytree()`` and/or ``-r``)

cc @kristjaneerik


stor/base.py
```diff
@@ -371,14 +373,17 @@ class FileSystemPath(Path):
...
+        Opens a path.
 
         Creates parent directory if it does not exist.
+
+        Raises:
+            ValueError: if ambiguous path - see ``stor.utils.validate_file_path`` for more.
         """
-        kwargs.pop('swift_upload_kwargs', None)
+        if kwargs.pop('swift_upload_kwargs', None):
+            warnings.warn('swift_upload_kwargs will be removed in stor 2.0')
```
krhaas: This doesn't seem discussed elsewhere in PR?

stor/obs.py
```diff
@@ -286,11 +288,17 @@ def __init__(self, pth, mode='r', encoding=None, **kwargs):
...
                 Python 3 only.
+        Raises:
+            ValueError: if the mode is not valid or the path is ambiguous (see ``stor.copy``)
         """
         if mode not in self._VALID_MODES:
             raise ValueError('invalid mode for file: %r' % mode)
         if six.PY2 and encoding:  # pragma: no cover
             raise TypeError('encoding not supported in Python 2')
+        # TODO (jtratner): in stor 2.0 remove this explicit check and the warning
+        if pth.endswith('/'):
```
krhaas: do you want the `and not self.ext` part from `is_ambiguous` ?

stor/swift.py
```diff
@@ -688,18 +688,16 @@ def open(self, mode='r', encoding=None, swift_upload_options=None):
                 ("r" or "rb") and writing ("w", "wb")
             encoding (str): text encoding to use. Defaults to
                 ``locale.getpreferredencoding(False)`` (Python 3 only)
-            swift_upload_options (dict): DEPRECATED (use `stor.settings.use()`
-                instead). A dictionary of arguments that will be
-                passed as keyword args to `SwiftPath.upload` if any writes
-                occur on the opened resource.
-
+            swift_upload_options (dict): DEPRECATED (use `stor.settings.use()` instead).
```
krhaas: `stor.settings.use()` already added correct?

stor/obs.py
krhaas: isn't call to `utils.validate_file_path` right below sufficient?

stor/utils.py
```diff
@@ -312,6 +313,52 @@ def is_writeable(path, swift_retry_options=None):
...
+    We can't do that kind of validation and don't want the outcome of a command to depend upon
+    whether there's a file with the same prefix or not. Thus, we raise an error if you try to use
+    `stor.copy()`. (to manipulate files without extensions, you can always use `stor.copytree()`.
+    """
+    import stor.settings
+
+    # we use a try/except so that it's easier for the reader to tell why the exception was
+    # generated.
+    try:
+        if not dest.name:
```
krhaas: is it slightly less explicit to rely on `dest.name` to verify no trailing slash rather than doing directly?

stor/utils.py
```diff
@@ -312,6 +313,52 @@ def is_writeable(path, swift_retry_options=None):
     return answer
 
 
+def validate_file_path(dest, _warn=False):
```
krhaas: nit: `dest` name implies this is always a check on destination path but it's called on both destination and source paths.

stor/utils.py
```diff
@@ -312,6 +313,52 @@ def is_writeable(path, swift_retry_options=None):
...
+
+    # we use a try/except so that it's easier for the reader to tell why the exception was
+    # generated.
+    try:
+        if not dest.name:
+            raise ValueError('file paths may not end with trailing slash')
+        if not dest.ext:
+            raise ValueError('file paths without extension are ambiguous')
+    except ValueError as e:
+        if ((_warn or is_filesystem_path(dest)) and
```
krhaas: should `is_filesystem_path` be moved earlier on since ambiguous paths are fine in that setting so no need to do any validations?

stor/utils.py
```diff
@@ -312,6 +313,52 @@ def is_writeable(path, swift_retry_options=None):
...
+    """
+    import stor.settings
+
+    # we use a try/except so that it's easier for the reader to tell why the exception was
+    # generated.
+    try:
+        if not dest.name:
+            raise ValueError('file paths may not end with trailing slash')
+        if not dest.ext:
+            raise ValueError('file paths without extension are ambiguous')
```
krhaas: requiring extensions for everything seems more restrictive that necessary and beyond what's required to resolve issue no?

pkaleta (approved):
adding empty review to remove this from my gh PRs list ;)

PR #64, opened 2018-01-20
Raise informative error on GETs from objects in Glacier

### LINKED ISSUE #38: Better error class for 403 on Glacier-backed files
Working with an S3 object in glacier, you get an `UnauthorizedError` if you try to access a file in glacier:

```
›› stor cat s3://bucket/key-in-glacier.json
UnauthorizedError: The operation is not valid for the object's storage class Bucket: bucket, Key: key-in-glacier.json
```

Similarly, you get an UnauthorizedError when using SwiftStack's Cloud Shunt towards a file in Glacier:

```
stor cat swift://A/C/some-file-in-glacier.json
UnauthorizedError: Object GET failed: https://swift.counsyl.com/v1/A/C/some-file-in-glacier.json 403 Forbidden   The operation is not valid for the object's storage class X-Trans-Id: <tid>

```

Given that sometimes 403s mean retries (and also we want to generate a useful error to user), I think we might want to say something like `ObjectInColdStorageError`:

```
ObjectInColdStorageError: < original error message> (Try a RESTORE operation on the object)
```

Here are some other alternatives for names:

`ObjectStorageClassError`, `GlacierGETError`, `UnrestoredObjectError`.

FYI:

Amazon calls it `InvalidObjectState` for their error code, but that name doesn't seem as directly actionable to the user.

```
HTTP/1.1 403 Forbidden
x-amz-request-id: CD4BD8A1310A11B3
x-amz-id-2: m9RDbQU0+RRBTjOUN1ChQ1eqMUnr9dv8b+KP6I2gHfRJZSTSrMCoRP8RtPRzX9mb
Content-Type: application/xml
Date: Mon, 12 Nov 2012 23:53:21 GMT
Server: AmazonS3
Content-Length: 231

<Error>
   <Code>InvalidObjectState</Code>
   <Message>The operation is not valid for the object's storage class</Message>
   <RequestId>9FEFFF118E15B86F</RequestId>
   <HostId>WVQ5kzhiT+oiUfDCOiOYv8W4Tk9eNcxWi/MK+hTS/av34Xy4rBU3zsavf0aaaaa</HostId>
</Error>
```

cc @krhaas @pkaleta @wesleykendall

jtratner:
Previously would raise UnauthorizedError, now we raise
ObjectInColdStorageError.  This is backwards-incompatible because
previously it does not subclass UnauthorizedError. (but this is a good
thing because we no longer auto-retry when something is in cold storage)

```
›› stor cat swift://AUTH_seq_upload_prod/X5WGWBCXY_161212_D00254_1027_A/.data_manifest.csv
ObjectInColdStorageError: Object GET failed: https://swift.counsyl.com/v1/AUTH_seq_upload_prod/X5WGWBCXY_161212_D00254_1027_A/.data_manifest.csv 403 Forbidden   The operation is not valid for the object's storage class X-Trans-Id: txe76bdb09dd8f44bf8750e-005a66619e
```

Sem-Ver: api-break

cc @pkaleta @wesleykendall @krhaas

Closes #38

### DIFF
--- a/stor/exceptions.py
+++ b/stor/exceptions.py
@@ -10,9 +10,18 @@ class RemoteError(Exception):
 
     The 'caught_exception' attribute of the exception must be examined in order
     to inspect the exception thrown by the swift or S3 service. A swift exception
-    can either be a ``SwiftError`` (thrown by ``swiftclient.service``) or a
-    ``ClientError`` (thrown by ``swiftclient.client``). A S3 exception can be
-    a ``ClientError`` (thrown by ``botocore.client``).
+
+    Swift has two types of caught exceptions:
+
+    * ``SwiftError`` (thrown by ``swiftclient.service``)
+    * ``ClientError`` (thrown by ``swiftclient.client``)
+
+    S3 raises:
+
+    * ``ClientError`` (thrown by ``botocore.client``).
+
+    Attributes:
+        caught_exception (Exception): the underlying exception that was raised from service.
     """
     def __init__(self, message, caught_exception=None):
         super(RemoteError, self).__init__(message)
@@ -22,6 +31,18 @@ class RemoteError(Exception):
 class NotFoundError(RemoteError):
     """Thrown when a 404 response is returned."""
     pass
+
+
+class ObjectInColdStorageError(RemoteError):
+    """Thrown when a 403 is returned from S3/SwiftStack because backing data is on Glacier.
+
+    Note:
+        We don't want to retry on this
+        one, because the object will always be in cold storage.
+        See AWS `S3 Rest API GET`_ spec for more details.
+
+    .. _S3 Rest API GET: https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html
+    """
 
 
 class UnauthorizedError(RemoteError):
--- a/stor/s3.py
+++ b/stor/s3.py
@@ -40,6 +40,7 @@ def _parse_s3_error(exc, **kwargs):
     """
     http_status = exc.response.get('ResponseMetadata', {}).get('HTTPStatusCode')
     msg = exc.response['Error'].get('Message', 'Unknown')
+    code = exc.response['Error'].get('Code')
 
     # Give some more info about the error's origins
     if 'Bucket' in kwargs:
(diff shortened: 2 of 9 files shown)


jtratner:
Closes #38

kyleabeauchamp:
LGTM

pkaleta:
looking

stor/exceptions.py
```diff
@@ -24,6 +24,15 @@ class NotFoundError(RemoteError):
     pass
 
 
+class ObjectInColdStorageError(RemoteError):
+    """Thrown when a 403 is returned from SwiftStack because backing data is on Glacier.
+
+    Note:
+        Currently only implemented for Swift. We don't want to retry on this
```
pkaleta: did you mean S3 too?

stor/s3.py
```diff
@@ -48,6 +49,8 @@ def _parse_s3_error(exc, **kwargs):
         msg += ', Key: ' + kwargs.get('Key')
 
     if http_status == 403:
+        if 'storage class' in msg and code == 'InvalidObjectState':
+            return exceptions.ObjectInColdStorageError(msg, exc)
```
pkaleta: Can you add a comment pointing to S3 doc that explains where this condition came from?

stor/swift.py
```diff
@@ -253,8 +253,11 @@ def _swiftclient_error_to_descriptive_exception(exc):
     if exc_headers and exc_headers.get('X-Trans-Id'):
         exc_str += ' X-Trans-Id: %s' % exc_headers['X-Trans-Id']
     if http_status == 403:
-        logger.error('unauthorized error in swift operation - %s', exc_str)
-        return UnauthorizedError(exc_str, exc)
+        if "storage class" in exc_str:
```
pkaleta: can you add a link to a doc that explains this

stor/s3.py
jtratner: see the ObjectInColdStorageError exception for details

stor/swift.py
jtratner: added a small comment, otherwise I suggest looking at the test case

PR #65, opened 2018-01-20
Add `to_url()` method on Path and `url` cli

jtratner:
to translate swift and s3 paths to HTTP paths.

(also fix up release notes because versions got off when we were not
auto-releasing)

=> @kristjaneerik @kyleabeauchamp @pkaleta @krhaas

### DIFF
--- a/stor/cli.py
+++ b/stor/cli.py
@@ -92,7 +92,7 @@ from stor import settings
 from stor import Path
 from stor import utils
 
-PRINT_CMDS = ('list', 'listdir', 'ls', 'cat', 'pwd', 'walkfiles', 'url')
+PRINT_CMDS = ('list', 'listdir', 'ls', 'cat', 'pwd', 'walkfiles', 'uri')
 SERVICES = ('s3', 'swift')
 
 ENV_FILE = os.path.expanduser('~/.stor-cli.env')
@@ -268,10 +268,10 @@ def get_path(pth, mode=None):
     return prefix / path_part.split(rel_part, depth)[depth].lstrip('/')
 
 
-def _to_url(path):
+def _as_uri(path):
     if stor.is_filesystem_path(path):
         raise ValueError('must be swift or s3 path')
-    return stor.Path(path).to_url()
+    return stor.Path(path).as_uri()
 
 
 def create_parser():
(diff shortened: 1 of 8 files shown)


jtratner:
Works for S3 - woo!

```
(.venv) jtratner on C13101jtratner in stor/on to-http-path
›› stor url s3://1000genomes/README_phase3_data_move_20150612                                                                      10:22 Jan19
https://1000genomes.s3.amazonaws.com/README_phase3_data_move_20150612
(.venv) jtratner on C13101jtratner in stor/on to-http-path
›› curl $(stor url s3://1000genomes/README_phase3_data_move_20150612)                                                              10:22 Jan19
The Phase 3 alignment and sequence data has been moved.

The alignments and sequence data can now be found under

ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/phase3/data

The variant calls are still available from

ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/release/20130502/
```

tophers42:
Nice! :+1:

stor/cli.py
```diff
@@ -268,6 +268,12 @@ def get_path(pth, mode=None):
     return prefix / path_part.split(rel_part, depth)[depth].lstrip('/')
 
 
+def _to_url(path):
+    if stor.is_filesystem_path(path):
+        raise ValueError('must be swift or s3 path')
```
tophers42: You should be able to make valid urls out of filesystem paths too, right?
```
curl file://localhost/<path>
```

kyleabeauchamp:
Overall LGTM once tests pass

stor/cli.py
jtratner: oh you know that's a really good point, then you can always do things like `open $(stor url $PATH)` will change

jtratner: okay, I'm actually only going to implement this for linux, because I'm not sure how it ought to work on windows :(

jtratner:
changing this to `as_uri()` because that's how pathlib calls it.

kristjaneerik:
Is a `from_url` planned?

jtratner:
### On file path URIs

there's some weird quirks with file paths, leaving that for a follow-up PR - see https://github.com/counsyl/stor/compare/master...jtratner:as-uri-file-path?expand=1 if you'd like to work on getting those tests to pass

### from_url

> `from_url`

would be cool, we could do it in a follow up PR - may leave as a good-first-PR target :)

General bikeshedding: is this actually better as `to_url` rather than `as_uri`, because it:

1. is actually an HTTP or browser-based URL (i.e., that your browser could open)
2. `s3://` is technically a URI too
3. makes it easier to see a `from_url` path

Those who care in this thread - go at it!

jtratner:
tests are passing, so would appreciate naming commentary :)

kristjaneerik:
I have a slight preference for the `to_url`-colored bikeshed 😸

PR #69, opened 2018-02-27
Remove deprecated top-level methods

jtratner:
1. remove `stor.listpath` and `stor.path`, both of which have been deprecated for a long time (and we're releasing v2.0 very soon anyways!)
2. update README.md to reflect new CLI options
3. change stor cp to no longer say "Alias for copy"

@kyleabeauchamp @pkaleta cc @wesleykendall

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,15 +27,17 @@ with a single, cross-compatible API.
 ## Quickstart - CLI
 
 ```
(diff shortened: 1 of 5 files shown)


jtratner:
thanks @pkaleta ! :)

PR #70, opened 2018-02-27
Make s3 test bucket configurable + append UUID to bucket name

jtratner:
Makes it possible to run s3 integration tests against your own S3 bucket + ensures that multiple test runs do not clobber each other. (previously they'd all run at $BUCKET/test, which is kinda silly)

@kyleabeauchamp @pkaleta


kyleabeauchamp:
LGTM

jtratner:
thanks @kyleabeauchamp ! :)

PR #71, opened 2018-03-05
Add S3Path.restore to restore objects from Glacier to S3

jtratner:
* Add ``stor.S3Path.restore(days, tier)`` to restore objects from Glacier to S3.
  Paired with this change are two specific exceptions related to S3 restores,
  ``RestoreAlreadyInProgressError`` (raised when restore has already started)
  and ``AlreadyRestoredError`` (raised when restoring object already in S3).
* Move ``ConflictError`` to ``stor.exceptions`` (still available under ``stor.swift``)

@pkaleta @kyleabeauchamp cc @wesleykendall

### DIFF
--- a/stor/exceptions.py
+++ b/stor/exceptions.py
@@ -34,7 +34,7 @@ class NotFoundError(RemoteError):
 
 
 class InvalidObjectStateError(RemoteError):
-    """Base class for 403 errors from S3/SwiftStack dealing with"""
+    """Base class for 403 errors from S3 dealing with storage classes."""
 
 
 class ObjectInColdStorageError(InvalidObjectStateError):
--- a/stor/exceptions.py
+++ b/stor/exceptions.py
@@ -33,7 +33,11 @@ class NotFoundError(RemoteError):
     pass
 
 
-class ObjectInColdStorageError(RemoteError):
+class InvalidObjectStateError(RemoteError):
+    """Base class for 403 errors from S3/SwiftStack dealing with"""
+
+
+class ObjectInColdStorageError(InvalidObjectStateError):
     """Thrown when a 403 is returned from S3/SwiftStack because backing data is on Glacier.
 
     Note:
@@ -43,6 +47,10 @@ class ObjectInColdStorageError(RemoteError):
 
     .. _S3 Rest API GET: https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html
     """
+
+
+class AlreadyRestoredError(InvalidObjectStateError):
+    """Thrown on attempt to restore object already not in Glacier"""
 
 
 class UnauthorizedError(RemoteError):
@@ -61,6 +69,22 @@ class UnauthorizedError(RemoteError):
 class UnavailableError(RemoteError):
     """Thrown when a 503 response is returned."""
     pass
+
+
+class ConflictError(RemoteError):
+    """Thrown when a 409 response is returned.
+
+    Notes:
+        * **Swift**: This error is thrown when deleting a container and
+          some object storage nodes report that the container
+          has objects while others don't.
+        * **S3**: Raised when attempting to restore object that's already being restored (as
+          RestoreAlreadyInProgressError). Possibly in other cases.
+    """
+
+
+class RestoreAlreadyInProgressError(ConflictError):
+    """Thrown when RestoreAlreadyInProgress on glacier restore"""
 
 
 class ConditionNotMetError(RemoteError):
--- a/stor/s3.py
+++ b/stor/s3.py
@@ -771,13 +771,16 @@ class S3Path(OBSPath):
                                                                  key=self.resource)
 
     def restore(self, tier='Bulk', days=10):
-        """Issue a restore command for the object from glacier.
+        """Issue a restore command for a single object from glacier.
 
         Args:
             tier (str, optional): restore speed (see Glacier docs for details)
             days (int, optional): number of days to keep data in S3 post-restore.
         Returns:
             dict or None: restore response from S3 client
+        Note:
+            Calling ``restore()`` on a directory will not work correctly. Only use this for single
+            objects! (you can always do ``[s.restore() for s in stor.list(<directory>)]``)
 
         Ignores RestoreAlreadyInProgressError and AlreadyRestoredError (note that you can't force
         S3 to do a faster restore once you've chosen a tier)
--- a/stor/s3.py
+++ b/stor/s3.py
@@ -774,10 +774,8 @@ class S3Path(OBSPath):
         """Issue a restore command for a single object from glacier.
 
         Args:
-            tier (str, optional): restore speed (see Glacier docs for details)
-            days (int, optional): number of days to keep data in S3 post-restore.
-        Returns:
-            dict or None: restore response from S3 client
+            tier (str, default 'Bulk'): restore speed (see Glacier docs for details)
+            days (int, default 10): number of days to keep data in S3 post-restore.
         Note:
             Calling ``restore()`` on a directory will not work correctly. Only use this for single
             objects! (you can always do ``[s.restore() for s in stor.list(<directory>)]``)
@@ -787,7 +785,7 @@ class S3Path(OBSPath):
         """
         valid_tiers = ('Standard', 'Bulk', 'Expedited')
         if tier not in valid_tiers:
-            raise TypeError('`tier` must be one of {}'.format(valid_tiers))
+            raise ValueError('`tier` must be one of {}'.format(valid_tiers))
         try:
             self._s3_client_call('restore_object',
                                  Bucket=self.bucket,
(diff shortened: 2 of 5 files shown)


jtratner:
cc @krhaas - centralizes error handling with S3 for restore tool

stor/exceptions.py
```diff
@@ -63,6 +71,22 @@ class UnavailableError(RemoteError):
     pass
 
 
+class ConflictError(RemoteError):
+    """Thrown when a 409 response is returned.
+
+    Notes:
```
krhaas: do you want these notes in the actual exception which inherits from `ConflictError` ?

jtratner: what do you mean?

stor/exceptions.py
```diff
@@ -33,7 +33,11 @@ class NotFoundError(RemoteError):
     pass
 
 
-class ObjectInColdStorageError(RemoteError):
+class InvalidObjectStateError(RemoteError):
+    """Base class for 403 errors from S3/SwiftStack dealing with"""
```
pkaleta: unfinished docstring

stor/s3.py
```diff
@@ -41,21 +41,25 @@ def _parse_s3_error(exc, **kwargs):
...
+                return exceptions.AlreadyRestoredError(msg, exc)
         return exceptions.UnauthorizedError(msg, exc)
-    elif http_status == 404:
+    if http_status == 404:
         return exceptions.NotFoundError(msg, exc)
-    elif http_status == 503:
+    if http_status == 503:
         return exceptions.UnavailableError(msg, exc)
+    if http_status == 409:
+        if 'Object restore is already in progress' in msg:
```
krhaas: I guess we just have to hope amazon keeps message the same. Or does this come from `boto`?

kyleabeauchamp:
So I made one pass through, but to do a "real" review the reviewer probably needs to be a bit more familiar with the glacier restore APIs.

stor/exceptions.py
krhaas: just to move the second part of comment comment to `RestoreAlreadyInProgressError` class

krhaas: looks like you already did though.

stor/s3.py
```diff
@@ -765,3 +769,34 @@ def to_url(self):
...
+        Args:
+            tier (str, optional): restore speed (see Glacier docs for details)
+            days (int, optional): number of days to keep data in S3 post-restore.
+        Returns:
+            dict or None: restore response from S3 client
+        Note:
+            Calling ``restore()`` on a directory will not work correctly. Only use this for single
+            objects! (you can always do ``[s.restore() for s in stor.list(<directory>)]``)
+
+        Ignores RestoreAlreadyInProgressError and AlreadyRestoredError (note that you can't force
```
krhaas: do we actually want to catch these errors?  What's wrong with raising to user?

stor/s3.py
```diff
@@ -765,3 +769,34 @@ def to_url(self):
         """Returns HTTP url for object (virtual host-style)"""
         return u'https://{bucket}.s3.amazonaws.com/{key}'.format(bucket=self.bucket,
                                                                  key=self.resource)
+
+    def restore(self, tier='Bulk', days=10):
+        """Issue a restore command for a single object from glacier.
+
+        Args:
+            tier (str, optional): restore speed (see Glacier docs for details)
```
pkaleta: add default

stor/s3.py
```diff
@@ -765,3 +769,34 @@ def to_url(self):
...
+            days (int, optional): number of days to keep data in S3 post-restore.
```
pkaleta: add default value

stor/s3.py
```diff
@@ -765,3 +769,34 @@ def to_url(self):
...
+        Returns:
+            dict or None: restore response from S3 client
+        Note:
+            Calling ``restore()`` on a directory will not work correctly. Only use this for single
```
pkaleta: can we detect this case and raise error?

stor/s3.py
```diff
@@ -765,3 +769,34 @@ def to_url(self):
...
                                                                  key=self.resource)
+
+    def restore(self, tier='Bulk', days=10):
+        """Issue a restore command for a single object from glacier.
+
+        Args:
+            tier (str, optional): restore speed (see Glacier docs for details)
+            days (int, optional): number of days to keep data in S3 post-restore.
+        Returns:
+            dict or None: restore response from S3 client
```
pkaleta: unclear when None is returned from just the docstring

stor/s3.py
```diff
@@ -765,3 +769,34 @@ def to_url(self):
...
+        Note:
+            Calling ``restore()`` on a directory will not work correctly. Only use this for single
+            objects! (you can always do ``[s.restore() for s in stor.list(<directory>)]``)
+
+        Ignores RestoreAlreadyInProgressError and AlreadyRestoredError (note that you can't force
+        S3 to do a faster restore once you've chosen a tier)
+        """
+        valid_tiers = ('Standard', 'Bulk', 'Expedited')
```
pkaleta: could add constants instead, up to you

stor/s3.py
```diff
@@ -765,3 +769,34 @@ def to_url(self):
...
+        if tier not in valid_tiers:
+            raise TypeError('`tier` must be one of {}'.format(valid_tiers))
```
pkaleta: TypeError is not the best exception type in this case: https://docs.python.org/2/library/exceptions.html#exceptions.TypeError

ValueError would be better.

stor/s3.py
jtratner: RestoreAlreadyInProgress happens when you do this:

```
path = stor.Path('s3://bucket/key/obj')
path.restore()
path.restore() # at this point Amazon returns a 409 Conflict, RestoreAlreadyInProgress error
```

AlreadyRestoredError:

```
path = stor.Path('s3://bucket/key/obj')
assert path.stat()['StorageClass'] != 'GLACIER'
path.restore()
```

in both cases, the user's intention (ensure this object is restored from Glacier to S3) is already satisfied (either because restore has already started or because it already is in S3).  In my view, bubbling up the error just makes the API more complicated to understand. (whereas without that error, you can just iterate over all objects on a bucket, hit restore on all of them and not need to know which errors to catch)

krhaas: maybe make an optional kwarg to raise or suppress?

krhaas: agree default should be to suppress. logger is good too.

stor/s3.py
jtratner: not really (aside from calling list() every single time you call restore which is both slow and costly) and tbh trying to do it just makes the API less useful for users (because either you try to string match to see things that look "prefix-like" and you block them or you miss cases where you should've seen it).

Right now this will raise an NotFoundError, like this:

```
In [2]: stor.Path('s3://some-bucket/my/key').restore()
---------------------------------------------------------------------------
NotFoundError                             Traceback (most recent call last)
<ipython-input-2-4275e295256d> in <module>()

/Users/jtratner/stor/stor/s3.py in restore(self, tier, days)
    794                                  Key=self.resource,
    795                                  RestoreRequest={'Days': days,
--> 796                                                  'GlacierJobParameters': {'Tier': tier}})
    797         except exceptions.RestoreAlreadyInProgressError:
    798             logger.debug('restore already started, not doing anything')

/Users/jtratner/stor/stor/s3.py in _s3_client_call(self, method_name, *args, **kwargs)
    200             return method(*args, **kwargs)
    201         except botocore_exceptions.ClientError as e:
--> 202             six.raise_from(_parse_s3_error(e, **kwargs), e)
    203
    204     def _get_s3_iterator(self, method_name, *args, **kwargs):

/usr/local/lib/python2.7/site-packages/six.pyc in raise_from(value, from_value)
    735 else:
    736     def raise_from(value, from_value):
--> 737         raise value
    738
    739

NotFoundError: The specified key does not exist. Bucket: some-bucket Key: my/key
```

I think that's good enough - we could potentially try to catch the `NotFoundError` and append ` - is it a prefix/directory? If so call restore() on each object individually` to the exception message

stor/s3.py
jtratner: did it, kinda duplicative with actual type signature tho

stor/s3.py
jtratner: if we need an optional kwarg at some future time we can add it. there's already logging at debug level

stor/s3.py
jtratner: again, kinda can take it as it comes... if amazon updates its API, we'll update this library :-/

stor/s3.py
pkaleta: What you have is good enough. 👍

stor/s3.py
jtratner: I didn't because I didn't want to add more items to the API... I'm open to a PR if we want to do that, just wasn't sure where you'd stick them...

PR #73, opened 2018-03-14
Convert README to restructured text

jtratner:
So it looks better on PyPI.

@pkaleta or @kyleabeauchamp if you got a sec


kyleabeauchamp:
LGTM

jtratner:
Will push this out along with another change I have coming up for stor that adds better list and other output :)

PR #76, opened 2018-04-20
xopen: open gzipped files transparently

kristjaneerik:
Here's an implementation of `stor.xopen` which works just like `stor.open` for regular files, but uses a layer of `gzip.GzipFile` for files ending in `.gz` such that you don't have to worry about compression, it happens behind the scenes.

Not sure if I did it correctly with respect to forcing the mode etc..

The portion of the test that runs `with stor.xopen(stor.join(self.drive, 'A/C/utf8_file_with_unicode.txt'), 'rb') as xfp:` doesn't work if the mode is `r`. I suspect it's something to do with the mocking rather than the implementation, since the code above it with mode `r` works. ¯\\\_(ツ)\_/¯

Suggestions for better approaches to testing appreciated!

Inspiration from https://github.com/marcelm/xopen. Would be great if we could just use that, but I'm not sure we can.


jtratner:
Looks nice, but have you tested this on python3? Try writing an emoji or
other Unicode into your file :-/ might blow up unfortunately

jtratner:
(If it works on python 3 then I’m super down and will review more in-depth!)

kristjaneerik:
I added an emoji to the test file.

All of these give `PASSED` as the result for the single test being run:

```bash
tox -e py27 -- stor/tests/test_swift.py::TestSwiftShared::test_xopen_gzip
tox -e py27 -- stor/tests/test_swift.py::TestSwiftShared::test_xopen_regular
tox -e py36 -- stor/tests/test_swift.py::TestSwiftShared::test_xopen_gzip
tox -e py36 -- stor/tests/test_swift.py::TestSwiftShared::test_xopen_regular
```

jtratner:
Awesome!

stor/base.py
```diff
@@ -272,6 +273,22 @@ def open(self, *args, **kwargs):
...
 
+    def xopen(self, mode='r', *args, **kwargs):
+        """Open a file-like object, transparently handling gzip compression.
+
+        See: :func:`open`
+        """
+        if self.endswith('.gz'):
+            if mode == 'r':
+                mode = 'rb'
+            if mode == 'w':
```
jtratner: I think these codepaths aren't actually tested (that's what is causing CI to fail).

Additionally, you need to pass the original mode to the gzip.GzipFile object, so perhaps you want something like:

```
if mode in ('r', 'rb'): fp_mode='rb'
if mode in ('w', 'wb'): fp_mode = 'wb'
fp = self.open(fp_mode, *args, **kwargs)
gzfp = gzip.GzipFile(mode=mode, fileobj=fp)
```

Additionally, GzipFile's docs say that it doesn't automatically close the underlying file object - which means that if you do something like this:

```
with stor.xopen('s3://unauthed-bucket/myfile.csv.gz') as fp:
    fp.write('somedata')
```
the exception will not bubble up to the user (I believe)

PR #79, opened 2018-07-26
Better progress logger for stor

jtratner:
Replace the existing progress logger with tqdm, a library that makes
elegant progress bars that calculate rate of transfer, time remaining,
and also look nice on the CLI (i.e., stays on one line and doesn't gum
up the entire works). A future enhancement would be to make the progress
logger configurable, but I figure for now just showing a better one is a
bonus.

### How this was tested:

1. Tested stor cp -r and rm -r with various combinatiosn
2. Ensure passes all unit tests.

I'm calling this a feature break: it does not break any existing APIs and the progress logger was never a public method exposed to users.

cc @kyleabeauchamp @wesleykendall @pkaleta @krhaas - if any of you have time to review that'd be awesome

### DIFF
--- a/stor/cli.py
+++ b/stor/cli.py
@@ -74,6 +74,7 @@ is not yet supported.
 """
 import argparse
 import copy
+from collections import OrderedDict
 from functools import partial
 import locale
 import logging
--- a/stor/s3.py
+++ b/stor/s3.py
@@ -113,63 +113,19 @@ def _get_s3_transfer(config=None):
 
 
 class S3DownloadLogger(utils.BaseProgressLogger):
-    def __init__(self, total_download_objects):
-        super(S3DownloadLogger, self).__init__(progress_logger)
-        self.total_download_objects = total_download_objects
-        self.downloaded_bytes = 0
+    logger = progress_logger
 
-    def update_progress(self, result):
-        """Tracks number of bytes downloaded."""
-        self.downloaded_bytes += (os.path.getsize(result['dest'])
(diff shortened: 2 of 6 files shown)


stor/cli.py
```diff
@@ -74,6 +74,7 @@
 """
 import argparse
 import copy
+from collections import OrderedDict
```
kyleabeauchamp: AFAIK this import is unused

stor/s3.py
```diff
@@ -113,63 +113,19 @@ def _get_s3_transfer(config=None):
...
-        return (
-            '%s/%s\t'
-            '%s\t'
-            '%0.2f MB\t'
-            '%0.2f MB/s'
-        ) % (self.num_results, self.total_download_objects, formatted_elapsed_time, mb, mb_s)
+    logger = progress_logger
+
+    def add_result(self, result):
+        self.update_progress(num_objects=1, addl_bytes=os.path.getszie(result['dest']))
```
kyleabeauchamp: Is this a typo?  `getszie`

jtratner: hmmm... maybe I didn't let the s3 version bubble up long enough..

jtratner:
This functionality is semi-covered by the integration tests (which call copytree and friends). I think I can test it  by checking for the calls to update_progress and/or grabbing the progress logger after running.

kyleabeauchamp:
A few typos to fix.  Overall idea SGTM.

PR #85, opened 2018-10-31
Implemented normpath and changed urllib to requests

anujkumar93:
Implemented custom normpath for DXPath, and added test cases
Changed urllib to requests in test_dx.py

### DIFF
--- a/stor/dx.py
+++ b/stor/dx.py
@@ -204,6 +204,18 @@ class DXPath(OBSPath):
         joined_resource = '/'.join(parts[1:]) if len(parts) > 1 else None
         return self.parts_class(joined_resource) if joined_resource else None
 
+    def normpath(self):
+        colon_pieces = self[len(self.drive):].split(':', 1)
+        project = colon_pieces[0]
+        resource = (colon_pieces[1] if len(colon_pieces) == 2 else '').lstrip('/')
+        normed_resource = self.path_module.normpath('/' + resource)[1:]
+        norm_pth = self.path_class(self.drive + project + ':/' + normed_resource)
+        if isinstance(norm_pth, DXCanonicalPath):
+            return self.path_class(self.drive + project + ':' + normed_resource)
+        if not norm_pth.resource:
+            norm_pth = utils.with_trailing_slash(norm_pth)
+        return norm_pth
+
     def dirname(self):
         """Returns directory name of path. Returns self if path is a project.
 
--- a/stor/dx.py
+++ b/stor/dx.py
@@ -156,6 +156,9 @@ class DXPath(OBSPath):
     def canonical_path(self):
         raise NotImplementedError
 
+    def normpath(self):
+        raise NotImplementedError
+
     @property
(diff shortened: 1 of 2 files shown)


stor/dx.py
```diff
@@ -204,6 +204,18 @@ def resource(self):
         joined_resource = '/'.join(parts[1:]) if len(parts) > 1 else None
         return self.parts_class(joined_resource) if joined_resource else None
 
+    def normpath(self):
+        colon_pieces = self[len(self.drive):].split(':', 1)
+        project = colon_pieces[0]
+        resource = (colon_pieces[1] if len(colon_pieces) == 2 else '').lstrip('/')
+        normed_resource = self.path_module.normpath('/' + resource)[1:]
+        norm_pth = self.path_class(self.drive + project + ':/' + normed_resource)
+        if isinstance(norm_pth, DXCanonicalPath):
```
jtratner: question:

why not just split out canonical path to a separate implementation for each type?  Then canonical path is just something like:

```
return self.path_class(self.drive + self.project + ':' + self.resource.lstrip('/'))
```

and for virtualpath you don't need to have that if statement anymore.

(i.e. `normed_resource = self.path_module.normpath(self.resource).lstrip('/')`, etc)

stor/dx.py
```diff
@@ -204,6 +204,18 @@ def resource(self):
...
+            return self.path_class(self.drive + project + ':' + normed_resource)
+        if not norm_pth.resource:
+            norm_pth = utils.with_trailing_slash(norm_pth)
```
jtratner: can we leave it as `dx://project:` rather than `dx://project:/` if there's no  resource?

stor/dx.py
anujkumar93: I actually started with that, but ran into a subtle issue. `stor.Path('dx://project-123456789012345678901234:/a/../file-123456789012345678901234')` is initialized to a DXVirtualPath, so if we call normpath() on this, it'll implement the virtual version and return `DXVirtualPath("dx://project-123456789012345678901234:/file-123456789012345678901234")`, whereas we would expect `DXCanonicalPath("dx://project-123456789012345678901234:file-123456789012345678901234")`.

This helps prevent that. The main catch here is that `norm_pth = self.path_class(self.drive + project + ':/' + normed_resource)` determines the initialization class. Because `stor.dx.DXVirtualPath('dx://project-123456789012345678901234:/file-123456789012345678901234')` is initialized to `DXCanonicalPath("dx://project-123456789012345678901234:/file-123456789012345678901234")`.

stor/dx.py
anujkumar93: Yes, for sure.

stor/dx.py
jtratner: seems like this that should return a DXCanonicalPath and you have a good test case for it :)

(also that's just a really weird example)

anujkumar93: Addressed.

PR #86, opened 2018-10-31
Stor-dx  Release Notes

anujkumar93:


docs/release_notes.rst
```diff
@@ -1,6 +1,12 @@
 Release Notes
 =============
 
+v2.1.4
```
jtratner: API break so needs to be v3.0.0

(also add a commit that says:

```
Sem-Ver: api-break
```

at the end to force this version change)

docs/release_notes.rst
```diff
@@ -1,6 +1,12 @@
...
+------
+* Add api and cli support for manipulating resources on DNAnexus platform.
```
jtratner:
```suggestion
* Add API and CLI support for manipulating resources on DNAnexus platform via the ``DXPath`` (i.e., the ``DXVirtualPath`` and ``DXCanonicalPath``) classes.
```

PR #87, opened 2018-10-31
Only allow bytes for write_object in Python 3

jtratner:
@anujkumar93 to review

Only allow bytes for write_object in Python 3. Motivation is to prevent latent programming bugs (since this is essentially required for Python 3 writing to buffers anyways).

I think this may break dxpy3 a bit, but right now it's already broken for Python 3 file object reading/writing...

### DIFF
--- a/stor/s3.py
+++ b/stor/s3.py
@@ -488,7 +488,8 @@ class S3Path(OBSPath):
         if not isinstance(content, bytes):  # pragma: no cover
             if six.PY2:
                 # bytes/unicode a little confused so allow it
-                warnings.warn('future versions of stor (and Python 3) will raise a TypeError if content is not bytes')
+                warnings.warn('Python 3 stor and a future Python 2 version of stor will raise a'
+                              ' TypeError if content is not bytes')
             else:
                 raise TypeError('write_object() expects bytes, not text data')
         mode = 'wb' if isinstance(content, bytes) else 'w'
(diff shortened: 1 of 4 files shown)


stor/s3.py
```diff
@@ -486,7 +486,11 @@ def write_object(self, content):
             content (bytes): raw bytes to write to OBS
         """
         if not isinstance(content, bytes):  # pragma: no cover
-            warnings.warn('future versions of stor will raise a TypeError if content is not bytes')
+            if six.PY2:
+                # bytes/unicode a little confused so allow it
+                warnings.warn('future versions of stor (and Python 3) will raise a TypeError if content is not bytes')
```
anujkumar93: `(and Python 2)`*. We are raising a TypeError already for Python3 now.

stor/swift.py
```diff
@@ -662,7 +662,11 @@ def write_object(self, content, **swift_upload_args):
                 `SwiftPath.upload`
         """
         if not isinstance(content, bytes):  # pragma: no cover
-            warnings.warn('future versions of stor will raise a TypeError if content is not bytes')
+            if six.PY2:
+                # bytes/unicode a little confused so allow it
+                warnings.warn('future versions of stor (and Python 3) will raise a TypeError if content is not bytes')
```
anujkumar93: same as above

PR #88, opened 2018-10-31
Simplify open arguments

jtratner:
Some additional API breaks to round out changes you already started @anujkumar93

* ``swift_upload_kwargs``  / ``swift_upload_options`` are no longer allowed in ``OBSFile()`` or ``stor.open()``
* The only permitted arguments/keyword arguments to ``open()`` are now ``mode`` and ``encoding``.

Haven't had chance to test but will test it in a final round of DX testing...

### DIFF
--- a/stor/base.py
+++ b/stor/base.py
@@ -262,7 +262,7 @@ class Path(text_type):
         """
         return self.path_class(self.path_module.join(self, *others))
 
-    def open(self, *args, **kwargs):
+    def open(self, mode=None, encoding=None):
         """Open a file-like object.
 
         The only cross-compatible arguments for this function are listed below.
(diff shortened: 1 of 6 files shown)


stor/base.py
```diff
@@ -371,17 +371,21 @@ class FileSystemPath(Path):
...
         Creates parent directory if it does not exist.
         """
-        kwargs.pop('swift_upload_kwargs', None)
         self.parent.makedirs_p()
-        return builtins.open(self, *args, **kwargs)
+        kwargs = {}
+        if mode:
+            kwargs['mode'] = mode
+        if encoding:
+            kwargs['encoding'] = encoding
```
anujkumar93: This can be DRYed up a bit:
```
kwargs = dict([(arg, eval(arg)) for arg in ('mode', 'encoding') if eval(arg)])
```

PR #89, opened 2018-10-31
Implementing wait_on_close upon open dx paths

anujkumar93:
wait_on_close is called from OBSFile after open(DXPath) is closed. This is because DX takes time to close the file and get it ready for download/read etc.

wait_on_close is added as a configurable parameter in stor settings.

### DIFF
--- a/stor/obs.py
+++ b/stor/obs.py
@@ -2,6 +2,7 @@ import locale
 import posixpath
 import sys
 
+import dxpy
 import six
 from swiftclient.service import SwiftError
 from swiftclient.service import SwiftUploadObject
@@ -9,6 +10,7 @@ from swiftclient.service import SwiftUploadObject
 from stor.base import Path
 from stor.posix import PosixPath
 from stor import utils
+import stor
 
 
 def _delegate_to_buffer(attr_name, valid_modes=None):
@@ -450,6 +452,22 @@ class OBSFile(object):
                 self.flush()
             self._buffer.close()
         self.closed = True
+        # we want to wait_on_close on DXPath only if no error was thrown while writing the file
+        if not any(sys.exc_info()):  # pragma: no cover
+            self.dx_close_helper()
+
+    def dx_close_helper(self):
+        if isinstance(self._path, stor.dx.DXPath):
+            wait_on_close = stor.settings.get()['dx']['wait_on_close']
+            if wait_on_close:
+                f = dxpy.DXFile(dxid=self._path.canonical_resource,
+                                project=self._path.canonical_project)
+                try:
+                    f.wait_on_close(wait_on_close)
+                # ignore timeout because we want client code to deal with it
+                except dxpy.DXError as e:  # pragma: no cover
+                    if 'timeout' not in str(e):
+                        raise
 
     def flush(self):
         """Flushes the write buffer to the OBS path (if it exists)"""
--- a/stor/obs.py
+++ b/stor/obs.py
@@ -454,20 +454,21 @@ class OBSFile(object):
         self.closed = True
         # we want to wait_on_close on DXPath only if no error was thrown while writing the file
         if not any(sys.exc_info()):  # pragma: no cover
-            self.dx_close_helper()
+            self._wait_on_close()
 
-    def dx_close_helper(self):
+    def _wait_on_close(self):
         if isinstance(self._path, stor.dx.DXPath):
             wait_on_close = stor.settings.get()['dx']['wait_on_close']
             if wait_on_close:
-                f = dxpy.DXFile(dxid=self._path.canonical_resource,
-                                project=self._path.canonical_project)
-                try:
-                    f.wait_on_close(wait_on_close)
-                # ignore timeout because we want client code to deal with it
-                except dxpy.DXError as e:  # pragma: no cover
-                    if 'timeout' not in str(e):
-                        raise
+                with stor.dx._wrap_dx_calls():
+                    f = dxpy.DXFile(dxid=self._path.canonical_resource,
+                                    project=self._path.canonical_project)
+                    try:
+                        f.wait_on_close(wait_on_close)
+                    # ignore timeout because we want client code to deal with it
+                    except dxpy.DXError as e:  # pragma: no cover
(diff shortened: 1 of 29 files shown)


stor/obs.py
```diff
@@ -450,6 +452,22 @@ def close(self):
                 self.flush()
             self._buffer.close()
         self.closed = True
+        # we want to wait_on_close on DXPath only if no error was thrown while writing the file
+        if not any(sys.exc_info()):  # pragma: no cover
+            self.dx_close_helper()
+
+    def dx_close_helper(self):
```
jtratner: naming wise a little funky:

probably better to make it private and just call

```
_wait_on_close
```

I'd also suggest changing signature to:

```
self._wait_on_close(20)
```

and then change: `wait_on_close` line as suggested below

stor/obs.py
```diff
@@ -450,6 +452,22 @@ def close(self):
...
+        if isinstance(self._path, stor.dx.DXPath):
+            wait_on_close = stor.settings.get()['dx']['wait_on_close']
```
jtratner:
```suggestion
            wait_on_close = wait_on_close if wait_on_close is not None else stor.settings.get()['dx']['wait_on_close']
```

stor/obs.py
```diff
@@ -450,6 +452,22 @@ def close(self):
...
+            if wait_on_close:
+                f = dxpy.DXFile(dxid=self._path.canonical_resource,
```
jtratner: btw - wouldn't it be more performant to just say `self._path.canonical_path.resource` and `self._path.canonical_path.project`

stor/obs.py
```diff
@@ -450,6 +452,22 @@ def close(self):
...
+                                project=self._path.canonical_project)
+                try:
+                    f.wait_on_close(wait_on_close)
+                # ignore timeout because we want client code to deal with it
+                except dxpy.DXError as e:  # pragma: no cover
```
jtratner: nit: `dxpy.DXFileError` according to docs

stor/obs.py
jtratner: does this need _wrap_dx_call too ?

stor/tests/cassettes_py3/TestLoginAuth/test_login_auth.yaml
```diff
@@ -239,18 +239,18 @@ interactions:
...
       Content-Length: ['41']
       Content-Type: [application/json; charset=utf-8]
-      Date: ['Tue, 30 Oct 2018 18:36:12 GMT']
+      Date: ['Wed, 31 Oct 2018 19:34:10 GMT']
       Server: [nginx]
       X-Content-Type-Options: [nosniff]
       X-Powered-By: [Express]
-      X-Request-ID: [1540924569244-887656]
+      X-Request-ID: [1541014448711-910820]
       X-Upgrade-Info: ['A recommended update (v0.267.0) is available for your client/SDK.
```
jtratner: random comment: does this need a freezegun and is that why it didn't work?

stor/tests/test_dx.py
```diff
@@ -278,12 +266,37 @@ def test_write_to_project_fail(self):
             with dx_p.open(mode='wb') as f:
                 f.write(b'data')
 
+    def test_write_w_settings_no_timeout(self):
+        self.setup_temporary_project()
+        dx_p = DXPath('dx://' + self.project + ':/temp_file')
+        with stor.settings.use({'dx': {'wait_on_close': 0}}):
```
jtratner: great test!!

stor/obs.py
anujkumar93: It is the same performance wise. canonical_path calls canonical_resource and canonical_project each once, and caches them.

stor/obs.py
anujkumar93: Manually made a file reach timeout error, as well as checked the code. It raises DXError, so probably inconsistency in doc.

Search for `_wait_on_close` here: http://autodoc.dnanexus.com/bindings/python/current/_modules/dxpy/bindings.html

stor/tests/cassettes_py3/TestLoginAuth/test_login_auth.yaml
anujkumar93: Nah, pretty sure that is okay. vcr matches only on uri and method, not the time of request or anything (hence, we don't need freezegun in general). I actually have no idea why it started failing in local. The only thing concerning the test is that we were talking about it :| . It seems to still work with Travis though, so I'm leaving it be for now.

stor/obs.py
anujkumar93: Changed name to _wait_on_close. And yes, it should have been wrapped in _wrap_dx_call. FIxed.

I am not so sure about changing the signature to 
```
    def _wait_on_close(self):
        if isinstance(self._path, stor.dx.DXPath):
            wait_on_close = wait_on_close if wait_on_close is not None else stor.settings.get()['dx']['wait_on_close']
```
and then call self._wait_on_close(20) from open, because then open would never work with settings. We are basically hardcoding 20 here. e.g. This won't take wait_on_close to be 0:
```
        with stor.settings.use({'dx': {'wait_on_close': 0}}):
            with dx_p.open(mode='wb') as obj:
                obj.write(b'hello world')
```

Let me know if I am getting this wrong.

jtratner: no I'm not suggesting you pass an explicit value to `_wait_on_close`, I'm just suggesting allowing specifying alternate wait period.

stor/obs.py
jtratner: gotcha

stor/tests/test_integration_dx.py
```diff
@@ -169,7 +158,7 @@ def test_over_100_files(self):
         self.assertEquals(120, len(self.test_dir.list(limit=120)))
         self.assertTrue(self.test_dir.isdir())
 
-        time.sleep(10)  # for uploaded files to close
+        time.sleep(20)  # for uploaded files to close
```
jtratner: why not just call `wait_on_close` on the last file? non-blocking tho

PR #91, opened 2018-11-05
doc removal of swiftfile

jtratner:
@anujkumar93


anujkumar93:
You probably need to include normal quotes: `'stor.swift.SwiftFile'`, else `make docs` won't work

PR #92, opened 2018-11-05
open fix for swift and s3

anujkumar93:
During fixing encoding issues for Python 3 and DX Paths, I overlooked that for swift and s3 paths, open is also delegated to OBSPath. Hence, the encoding check should only be for DX Paths in particular.

### DIFF
--- a/stor/obs.py
+++ b/stor/obs.py
@@ -166,7 +166,8 @@ class OBSPath(Path):
             DNAnexusError: A dxpy client error occured.
             RemoteError: A s3 client error occurred.
         """
-        if six.PY3 and encoding and encoding not in ('utf-8', 'utf8'):  # pragma: no cover
+        if six.PY3 and encoding and encoding not in ('utf-8', 'utf8') and \
+                isinstance(self, stor.dx.DXPath):  # pragma: no cover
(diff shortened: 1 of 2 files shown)


anujkumar93:
@jtratner . Also, when I split the modules, this will be taken care of as I will subclass open functions separately, but this seemed like an important fix which shouldn't have to wait until then.

jtratner:
needs a test case

anujkumar93:
The reason the tests didn't catch this was because we test 'utf-16' only in integration tests. Am adding the following as a unit test case (normally we don't check for Py2 etc in unit tests but have to here)
```
@unittest.skipIf(six.PY2, 'encoding only tested for Python3')
    def test_read_success_utf16(self):  # pragma: no cover
        self.mock_swift_conn.get_object.return_value = ('header', 'data'.encode('utf-16'))
        swift_p = SwiftPath('swift://tenant/container/obj')
        self.assertEquals(swift_p.read_object(), 'data'.encode('utf-16'))
        self.assertEquals(swift_p.open(encoding='utf16').read(), 'data'.encode('utf-16'))
```


At what point are the integration tests run? Is it when we merge dnanexus branch into master?

PR #93, opened 2018-11-05
made wait_on_close default to 0 and added docs

anujkumar93:
@jtratner

### DIFF
--- a/stor/tests/test_dx.py
+++ b/stor/tests/test_dx.py
@@ -189,6 +189,9 @@ class TestRename(DXTestCase):
 
 
 class TestOpen(DXTestCase):
+    # we need to give enough time for file to enter `closed` state so we can read it
+    DX_WAIT_SETTINGS = {'dx': {'wait_on_close': 20}}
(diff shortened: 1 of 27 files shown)


stor/tests/test_dx.py
```diff
@@ -254,17 +255,19 @@ def test_write_multiple_wo_context_manager(self):
         obj = dx_p.open(mode='wb')
         obj.write(b'hello')
         obj.write(b' world')
-        obj.close()
+        with stor.settings.use({'dx': {'wait_on_close': 20}}):
```
jtratner: how does this test what you're looking for? Or is the goal here just to give 20 seconds so we can be sure file is done?

If the latter, perhaps:

```
with stor.settings.us(DX_WAIT_SETTINGS):
    # do stuff
```

and earlier
```
# we need to give enough time for file to enter `closed` state so we can read it
DX_WAIT_SETTINGS = {'dx': {'wait_on_close': 20}}
```

PR #97, opened 2018-11-16
Split stor into 4 packages- stor, stor_swift, stor_s3, stor_dx

anujkumar93:
This is follow up to PR https://github.com/counsyl/stor/pull/96 . This creates each module into its own package (with different setup.py). The base stor picks up the 'extra' packages from 'stor.providers' entry points.

Things that have changed that need review:

- copy/copytree : code significantly changed, logic hasn't
- cli : had to hardcode drives in places and behavior of _convert_swiftstack has changed
- a few tests - added init tests for packages (find_cls_for_pth method)
- docs
- tox.ini - root package is not installed (by setting skipsdist=True) but apparently still needs an empty tox.ini to be present. the packages are installed in commands, and pytest cmd modified.
- makefile : among other necessary things, travis-test was forwarded to test
- requirements.txt : all the packages have requirements.txt but only base stor has requirements-setup.txt, requirements-dev.txt, etc


stor/stor/utils.py
```diff
@@ -410,19 +279,13 @@ def copy(source, dest, **kwargs):
             ...
             ValueError: OBS destination must be file with extension or directory with slash
     """
-    from stor import Path
     from stor.obs import OBSUploadObject
-
-    source = Path(source)
-    dest = Path(dest)
+    source = stor.Path(source)
+    dest = stor.Path(dest)
```
jtratner: It seems like this would become a lot simpler if we made these into shared methods rather than into utils.

for the POSIX implementation, you always know that source is a file or directory, so the other checks are not required.

for the OBS implementation, for Swift and S3, non-POSIX dest is not supported; for DX, only other DX is supported. (I'm thinking you could write one method along with a `supports_server_side_copy` flag that's True for DX and False for Swift and S3)

stor_dx/stor_dx/dx.py
```diff
@@ -884,8 +956,8 @@ def list(self,
         ))
         if not results or not results[0]:  # when results == [[]]
             results = []
-        utils.validate_condition(condition)
-        utils.check_condition(condition, results)
+        stor_utils.validate_condition(condition)
+        stor_utils.check_condition(condition, results)
```
jtratner: perhaps `core_utils` would be a better name for this?

anujkumar93:
Changes for https://github.com/counsyl/stor/pull/96#discussion_r236411957 will be done in this PR.

stor/stor/utils.py
anujkumar93: Done as part of https://github.com/counsyl/stor/pull/98

PR #98, opened 2018-11-29
Splitting copy/copytree across packages and changing its signature

anujkumar93:
New structure for copy/copytree

### DIFF
--- a/stor/base.py
+++ b/stor/base.py
@@ -87,126 +87,14 @@ class Path(text_type):
         compatible path module for path operations"""
         return isinstance(other, Path) and other.path_module != self.path_module
 
-    def copy(self, dest, **swift_retry_options):
-        """Copies a source file to a destination file.
-
-        Note that this utility can be called from either OBS, posix, or
-        windows paths created with ``stor.Path``.
-
-        Args:
-            self (path|str): The source directory to copy from
-            dest (path|str): The destination file or directory.
-            swift_retry_options (dict): Optional retry arguments to use for swift
-                upload or download. View the
-                `swift module-level documentation <swiftretry>` for more
-                information on retry arguments
-
-        Examples:
-            Copying a swift file to a local path behaves as follows::
-
-                >>> import stor
-                >>> swift_p = 'swift://tenant/container/dir/file.txt'
-                >>> # file.txt will be copied to other_dir/other_file.txt
-                >>> stor.copy(swift_p, 'other_dir/other_file.txt')
-
-            Copying from a local path to swift behaves as follows::
-
-                >>> from stor import Path
-                >>> local_p = Path('my/local/file.txt')
-                >>> # File will be uploaded to swift://tenant/container/dir/my_file.txt
-                >>> local_p.copy('swift://tenant/container/dir/')
-
-            Because of the ambiguity in whether a remote target is a file or directory, copy()
-            will error on ambiguous paths.
-
-                >>> local_p.copy('swift://tenant/container/dir')
-                Traceback (most recent call last):
-                ...
-                ValueError: OBS destination must be file with extension or directory with slash
-        """
-        from stor.obs import OBSUploadObject
-        dest = Path(dest)
-
-        if utils.is_filesystem_path(dest):
-            dest.parent.makedirs_p()
-            shutil.copy(self, dest)
-        else:
-            dest_file = dest if not dest.endswith('/') else dest / self.name
-            dest_obj_name = Path(dest_file.parent.resource or '') / dest_file.name
-            upload_obj = OBSUploadObject(self, dest_obj_name)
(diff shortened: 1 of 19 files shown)


stor_swift/tests/test_cli_swift.py
```diff
@@ -22,7 +22,7 @@ def setUp(self):
         patcher.start()
 
     @contextlib.contextmanager
-    def assertOutputMatches(self, exit_status=None, stdout='', stderr=''):
+    def assertOutputMatches(self, exit_status=None, stdout='', stderr=''):  # pragma: no cover
```
jtratner: add a test for this brah

anujkumar93: It is just a helper function used by tests in test_cli_swift. Since some stuff was moved here from test_cli.py, I also had to move this function here. This currently resembles the same func from test_cli.py but I can scope it to the facets that are used by swift tests:
```
    def assertOutputMatches(self, exit_status=None, stdout='', stderr=''):
        patch = mock.patch('sys.stderr', new=six.StringIO())
        self.addCleanup(patch.stop)
        patch.start()
        try:
            yield
        except SystemExit as e:
            self.assertEquals(e.code, int(exit_status))
        self.assertEquals(sys.stdout.getvalue(), '', 'stdout')
        self.assertRegexpMatches(sys.stderr.getvalue(), stderr, 'stderr')
```
Don't have to use no pragma then. Is this ok?

jtratner: the reason you need pragma no cover is because you're missing a test of a test.  Just need something like:

```
with self.assertRaises(AssertionError):
    with self.assertOutputMatches(exit_code=10):
        self.run('stor blahrg')
with self.assertRaises(AssertionError):
    with self.assertOutputMatches(stdout='mystdout'):
        self.run('stor -h')
```

it's just confirming that your matcher can actually fail :)

stor/base.py
```diff
@@ -86,8 +87,126 @@ def _has_incompatible_path_module(self, other):
         compatible path module for path operations"""
         return isinstance(other, Path) and other.path_module != self.path_module
 
-    copy = utils.copy
-    copytree = utils.copytree
+    def copy(self, dest, **swift_retry_options):
```
jtratner: I think this method should be on FileSystemPath, not the base path class.

Also let's get rid of swift_retry_options :)

stor/base.py
```diff
@@ -86,8 +87,126 @@ def _has_incompatible_path_module(self, other):
...
+            dest.parent.makedirs_p()
+            shutil.copy(self, dest)
+        else:
+            dest_file = dest if not dest.endswith('/') else dest / self.name
+            dest_obj_name = Path(dest_file.parent.resource or '') / dest_file.name
+            upload_obj = OBSUploadObject(self, dest_obj_name)
+            dest_file._copy_upload([upload_obj],
+                                   **swift_retry_options)
+
+    def copytree(self, dest, use_manifest=False, headers=None,
```
jtratner: same point here - move to filesystem path

stor/obs.py
```diff
@@ -73,8 +73,124 @@ def __init__(self, pth):
             raise ValueError('path must have %s (got %r)' % (self.drive, pth))
         return super(OBSPath, self).__init__(pth)
 
-    copy = utils.copy
-    copytree = utils.copytree
+    def copy(self, dest, **swift_retry_options):
```
jtratner: remove swift_retry_options

stor/obs.py
```diff
@@ -73,8 +73,124 @@ def __init__(self, pth):
...
+        """Copies a source file to a destination file.
+
+        Note that this utility can be called from either OBS, posix, or
+        windows paths created with ``stor.Path``.
+
+        Args:
+            self (path|str): The source directory to copy from
```
jtratner:
```suggestion
            self (Path|str): The source directory to copy from
```

stor_swift/tests/test_cli_swift.py
anujkumar93: Agreed, changed

stor/base.py
anujkumar93: I moved it to FileSystemPath. Are we sure we want to get rid of swift_retry_options? It's a big api break. And will have changes both upstream (packages that invoke `stor.copy(..., ..., swift_retry_options=...)` and downstream (upload, download_object, download all have declarations like `def upload(self, to_upload, **kwargs):`, also tests will change). Let me know.

stor/base.py
anujkumar93: Moved

anujkumar93:
This won't disallow server-side copy for DNAnexus. Because DXPath defines its own copy function and defaults to copy of OBSPath only if it's not server-side copy

jtratner:
lgtm

PR #104, opened 2019-01-15
Fix for normpath for DXPath

anujkumar93:
@jtratner @pkaleta 

This PR fixes the normpath bug in DXVirtualPath and DXCanonicalPath

## Background
Stor had a bug for DXPaths, wherein paths like DXPath('dx://project:/').normpath() errored with `TypeError: cannot concatenate 'str' and 'NoneType' objects`. This is because for such paths, the resource turns out to be None, which is then attempted to add it to a string '/' in normpath.

## Changes
- Fixed the normpath function in DXVirtualPath and DXCanonicalPath through `self.resource or ''`
- Added tests for normpath

## Tested via:
- Test suite was run successfully

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1,6 +1,12 @@
 Release Notes
 =============
 
+v3.0.1
+------
+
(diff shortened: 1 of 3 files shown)


jtratner:
please add release note for this patch release

PR #107, opened 2019-02-20
Posix to DX copy should delete dest file, if already existing

anujkumar93:
@jtratner @pkaleta 

Fixes/implements [BIOINF-1425](https://jira.counsyl.com/browse/BIOINF-1425) wherein we delete the destination file on DNAnexus during a posix to dx copy function

## Background

Currently, the 'delete existing dest file if it already exists' functionality happens only on DX to DX stor.copy, and during stor.open. It does not happen for stor.copy on posix to dx paths. This needed to be fixed.

## Changes

- Added check in utils.py (utils.copy) to remove dest file if it already exists
- Changed DXPath.upload to only print out a skip message if any of the provided files already exist on dnanexus platform.
- Added tests for upload and re-recorded posix_to_dx copy cassettes

## Tested via:

- The unit and integration test suite was run successfully


stor/tests/test_dx.py
```diff
@@ -1177,15 +1180,16 @@ def test_posix_to_dx_fail(self):
...
         self.setup_posix_files(['/rand/file.txt'])
-        dx_p = DXPath('dx://' + self.project + ':/temp_folder')
+        dx_folder_p = DXPath('dx://' + self.project + ':/temp_folder')
         posix_p = Path('./{test_folder}/{path}'.format(
             test_folder=self.project, path='rand/file.txt'))
-        with pytest.raises(exceptions.TargetExistsError, match='will not cause duplicate file'):
-            posix_p.copy(dx_p)
+        posix_p.copy(dx_folder_p)
+        dx_p = DXPath('dx://' + self.project + ':/temp_folder/file.txt')
+        self.assertTrue(dx_p.exists())
```
pkaleta: Should you also assert that the contents are from the correct file? I.e. that it actually got overwritten in DX?

stor/dx.py
```diff
@@ -779,11 +779,11 @@ def upload(self, to_upload, **kwargs):
 
             if upload_obj.source.isfile():
                 dest_is_file = dest_file.isfile()
-                if dest_is_file:
-                    raise stor_exceptions.TargetExistsError(
+                if dest_is_file:  # only occurs if upload is called directly with existing objects
+                    print(stor_exceptions.TargetExistsError(
```
pkaleta: Why do you create an instance of `TargetExistsError` here? Should this be a logger statement instead of print?

stor/tests/test_dx.py
anujkumar93: Yes, added that here and in TestUpload.

stor/dx.py
anujkumar93: yep that makes sense. changed to logger.warning. removed the TargetExistsError.

PR #110, opened 2019-03-22
Fix stor.exists for DXCanonicalPaths

anujkumar93:
@jtratner CC @pkaleta 

This PR fixes stor.exists for DXCanonicalPaths. Due to the slightly vague nature of dxpy's describe method, stor.exists gave wrong results for DXCanonicalPaths. Fixed that, and added tests.

### DIFF
--- a/stor/tests/test_dx.py
+++ b/stor/tests/test_dx.py
@@ -811,6 +811,51 @@ class TestExists(DXTestCase):
         result = dx_p.exists()
         self.assertFalse(result)
 
+    def test_true_canonical_project(self):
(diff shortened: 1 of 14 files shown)


stor/tests/test_dx.py
```diff
@@ -811,6 +811,51 @@ def test_project_does_not_exist(self):
         result = dx_p.exists()
         self.assertFalse(result)
 
+    def test_true_canonical_project(self):
```
jtratner: nice tests! :)

PR #111, opened 2019-03-22
updating dxpy3 to dxpy

anujkumar93:
@jtratner CC @pkaleta 

Since DNAnexus has integrated dxpy3 into dxpy, makes sense to get rid of dxpy3 from stor.

Passed unit and integration tests.


anujkumar93:
I had included the release notes along with other small PRs in this single PR: https://github.com/counsyl/stor/pull/114/ . But since the other ones still need reviewing, adding the release notes separately here. I think this is good to merge after this.

PR #113, opened 2019-03-24
Enable empty path dx:// for DXPath initialization

anujkumar93:
@jtratner CC @pkaleta 

## Background

This PR enables initialization of dx:// as a DXPath. We want to allow initialization of `dx://` to be able to call listdir on it and see the projects the user has access to. `stor ls dx://` will list such projects.

## Changes

- Changed DXPath initialization logic in utils to provision for dx://
- Handled logic in all DXPath function to allow for dx://
- Added tests for dx:// for all DXPath functions

### Other unrelated tiny changes in this PR:
- Added splitpath in DXPath. Had moved it earlier to individual subclasses but needed a declaration in DXPath too
- Added my name to AUTHORS :)

### DIFF
--- a/stor/dx.py
+++ b/stor/dx.py
@@ -100,7 +100,7 @@ class DXPath(OBSPath):
     def __new__(cls, path):
         """Custom __new__ method so that the validation checks happen during creation
 
-        This ensures invalid dx paths like DXPath('dx://') are never initialized
+        This ensures that dx paths are initialized to the correct subclass
         """
         return super(DXPath, cls).__new__(Path, path)
 
@@ -138,25 +138,34 @@ class DXPath(OBSPath):
         for prop in ('canonical_project', 'canonical_resource', 'virtual_path', 'virtual_project'):
             self.__dict__.pop(prop, None)
 
+    @property
     def virtual_project(self):
         raise NotImplementedError
 
+    @property
     def virtual_resource(self):
         raise NotImplementedError
 
+    @property
     def virtual_path(self):
         raise NotImplementedError
 
+    @property
     def canonical_project(self):
         raise NotImplementedError
 
+    @property
     def canonical_resource(self):
         raise NotImplementedError
 
+    @property
     def canonical_path(self):
         raise NotImplementedError
 
     def normpath(self):
+        raise NotImplementedError
+
+    def splitpath(self):
         raise NotImplementedError
 
     @property
@@ -185,7 +194,7 @@ class DXPath(OBSPath):
             NotFoundError: The path could not be resolved to a file
         """
         if not self.resource:
-            raise ValueError('DX Projects cannot have a temporary download url')
+            raise ValueError('DX Paths need path resource to have a temporary download url')
         with _wrap_dx_calls():
             file_handler = dxpy.DXFile(dxid=self.canonical_resource,
                                        project=self.canonical_project)
@@ -231,8 +240,7 @@ class DXPath(OBSPath):
 
     @property
     def name(self):
-        """File or folder name of the path. Empty string for projects or folders
-        with trailing slash.
+        """File or folder name of the path. Empty string for paths with trailing slash.
 
         Makes no API calls to server, canonical paths are treated normally, and the basename
         of the path is returned.
@@ -306,6 +314,8 @@ class DXPath(OBSPath):
         Returns:
             bool: True if path is an existing folder path or project
         """
+        if not self.project:  # for empty path dx://
+            return False
         if not self.resource and self.exists():  # path could be a project
             return True
         # or path could be a folder
@@ -340,7 +350,7 @@ class DXPath(OBSPath):
             NotFoundError: When path cannot be resolved to a file.
         """
         if not self.resource:
-            raise ValueError('Projects cannot be renamed')
+            raise ValueError('Paths without resource cannot be renamed')
         if new_name == self.name:
             return
         file_handler = dxpy.DXFile(dxid=self.canonical_resource,
@@ -696,6 +706,9 @@ class DXPath(OBSPath):
     def download_objects(self,
                          dest,
                          objects):
+        if not self.project:
+            raise ValueError('Cannot call download_objects on empty path ({})'.format(self))
+
         def is_parent_dir(possible_parent, possible_child):
             """Checks if possible_child is a sub-path of possible_parent"""
             if not possible_parent.resource:
@@ -809,7 +822,7 @@ class DXPath(OBSPath):
             bytes: the raw bytes from the object on DX.
         """
         if not self.resource:
-            raise ValueError('Can only read_object() on a file path, not a project')
+            raise ValueError('Can only call read_object() on a file path')
         file_handler = dxpy.DXFile(dxid=self.canonical_resource,
                                    project=self.canonical_project)
         with _wrap_dx_calls():
@@ -832,7 +845,7 @@ class DXPath(OBSPath):
                 `DXPath.upload`
         """
         if not self.resource:
-            raise ValueError('Cannot write to project. Please provide a file path')
+            raise ValueError('Cannot write to provided path. Please provide a valid file path')
         if not isinstance(content, bytes):  # pragma: no cover
                 # bytes/unicode a little confused so allow it
                 warnings.warn('Python 3 stor and a future Python 2 version of stor will raise a'
@@ -923,8 +936,30 @@ class DXPath(OBSPath):
             classname=classname
         )
 
+    def _listdir_empty_path(self, canonicalize=False):
+        ans_list = []
+        kwargs = {
+            'describe': {'fields': {'name': True}},
+            # `dx select` uses CONTRIBUTE default, but we want to show all view level projects
+            'level': 'VIEW',
+            # do not list public projects where user isn't specifically invited
+            'explicit_perms': True
+        }
+        projects = list(dxpy.find_projects(**kwargs))
+        for project in projects:
+            if canonicalize:
+                ans_list.append(DXCanonicalPath('{drive}{project}:'.format(
+                    drive=self.drive, project=project['id'])))
+            else:
+                ans_list.append(DXVirtualPath('{drive}{proj_name}:'.format(
+                    drive=self.drive,
+                    proj_name=project['describe']['name']))
+                )
+        return ans_list
+
     def listdir(self, only='all', canonicalize=False):
         """List the path as a dir, returning top-level directories and files.
+        If listing an empty path, lists the projects a user has access to.
 
         Args:
             canonicalize (bool, default False): if True, return canonical paths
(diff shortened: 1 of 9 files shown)


stor/dx.py
```diff
@@ -923,8 +936,30 @@ def list_iter(self,
...
+        ans_list = []
+        kwargs = {
+            'describe': {'fields': {'name': True}},
+            # `dx select` uses CONTRIBUTE default, but we want to show all view level projects
+            'level': 'VIEW',
+            # do not list public projects where user isn't specifically invited
+            'explicit_perms': True
+        }
+        projects = list(dxpy.find_projects(**kwargs))
+        for project in projects:
```
jtratner: Is there a reason to do a for-loop vs.:

```
if canonicalize:
    return [DXCanonicalPath('{drive}{project}:'.format(drive=self.drive,project=project['id']) for project in projects]
else:
   return [DXVirtualPath('...') for project in projects]
```

stor/dx.py
```diff
@@ -967,13 +1006,30 @@ def listdir(self, only='all', canonicalize=False):
...
         Returns:
             Iterable[DXPath]: Iterates over listed files directly within the resource
         """
+        if not self.project:  # list projects for empty path
+            kwargs = {
+                'describe': {'fields': {'name': True}},
+                'level': 'VIEW',
+                'explicit_perms': True
+            }
+            projects_generator = dxpy.find_projects(**kwargs)
```
jtratner: will this find public projects too?

stor/dx.py
```diff
@@ -967,13 +1006,30 @@ def listdir(self, only='all', canonicalize=False):
 
     def listdir_iter(self, canonicalize=False):
         """Iterate the path as a dir, returning top-level directories and files.
+        If listing an empty path, lists the projects a user has access to.
```
jtratner: needs more context (especially in docs) about which projects are seen when you do ls on `dx://`

stor/tests/cassettes_py2/TestCopyTree/test_empty_path.yaml
```diff
@@ -0,0 +1,139 @@
...
+      - application/json
+      DNAnexus-API:
+      - 1.0.0
+      User-Agent:
+      - dxpy/0.278.0 (Darwin-18.2.0-x86_64-i386-64bit)
+    method: POST
+    uri: https://api.dnanexus.com/system/findProjects
+  response:
+    body:
+      string: !!python/unicode '{"results":[{"id":"project-FXFPP6801YyXv3Q32v06xKyz","level":"ADMINISTER","permissionSources":["user-akumar_counsyl"],"public":false}],"next":null}'
```
jtratner: why does this only end up returning one project?  Also shouldn't this be returning project name too?

stor/dx.py
```diff
@@ -832,7 +845,7 @@ def write_object(self, content, **kwargs):
                 `DXPath.upload`
         """
         if not self.resource:
-            raise ValueError('Cannot write to project. Please provide a file path')
+            raise ValueError('Cannot write to provided path. Please provide a valid file path')
```
jtratner:
```
Must include a valid project and path for writing. Please provide a file path
```

might be clearer

stor/dx.py
```diff
@@ -185,7 +194,7 @@ def temp_url(self, lifetime=300, filename=None):
             NotFoundError: The path could not be resolved to a file
         """
         if not self.resource:
-            raise ValueError('DX Projects cannot have a temporary download url')
+            raise ValueError('DX Paths need path resource to have a temporary download url')
```
jtratner:
```
Can only generate download URLs for file paths, not projects. Please provide a valid file path.
```

stor/dx.py
```diff
@@ -340,7 +350,7 @@ def _rename(self, new_name):
             NotFoundError: When path cannot be resolved to a file.
         """
         if not self.resource:
-            raise ValueError('Projects cannot be renamed')
+            raise ValueError('Paths without resource cannot be renamed')
```
jtratner: I think the original message is actually fine (and clearer for 90% of all cases).  or do an if statement:

```
if not self.resource:
   raise ValueError('Projects cannot be renamed' if self.project else 'Cannot rename dx root')
```

stor/dx.py
```diff
@@ -809,7 +822,7 @@ def read_object(self):
             bytes: the raw bytes from the object on DX.
         """
         if not self.resource:
-            raise ValueError('Can only read_object() on a file path, not a project')
+            raise ValueError('Can only call read_object() on a file path')
```
jtratner: might be nice to do something like `+ (', not a project' if self.project else '')` here

stor/dx.py
```diff
@@ -1006,6 +1062,9 @@ def walkfiles(self,
         Returns:
              Iter[DXPath]: Iterates over listed files that match an optional pattern.
         """
+        if not self.project:  # for empty paths
+            raise ValueError('list, walkfiles and glob are not supported on empty paths. '
```
jtratner: why not allow `list()` on root?

stor/tests/test_dx_path_compat.py
```diff
@@ -17,8 +17,12 @@ def test_construction_from_none(self):
...
 
     def test_construction_from_no_project(self):
-        with pytest.raises(ValueError, match='Project is required to construct a DXPath'):
-            DXPath('dx://')
+        dx_p = DXPath('dx://')
+        self.assertEqual(dx_p, 'dx://')
+
+    def test_construction_no_project_fail(self):
+        with self.assertRaises(ValueError):
+            DXPath('dx:///folder')
```
jtratner: can you add a test with a value error on something like ```DXPath('dx://myproject/myfolder')``` just for clarity?

stor/tests/test_dx.py
```diff
@@ -1386,6 +1502,17 @@ def test_dx_canonical_to_dx_file(self):
 
 
 class TestCopyTree(DXTestCase):
+    def test_empty_path(self):
+        self.setup_temporary_project()
+        dx_p = DXPath('dx://')
+        dest = DXPath('dx://' + self.project)
+        with pytest.raises(exceptions.NotFoundError, match='No project or directory was found'):
```
jtratner: to me, this error message reads as if the destination doesn't *exist*.  Instead what you're trying to say is more like:

```
Cannot copy to or from root (dx://). Must specify project and/or directory or file.
```

stor/tests/test_dx.py
```diff
@@ -1386,6 +1502,17 @@ def test_dx_canonical_to_dx_file(self):
...
+            dx_p.copytree(dest)
+        with pytest.raises(NotImplementedError):
+            dx_p._clonetree(dest)
```
jtratner: seems kinda unnecessary - alway why do d`NotImplementedError` here?

stor/tests/test_dx.py
```diff
@@ -769,6 +844,11 @@ def test_stat_canonical_resource(self):
 
 
 class TestExists(DXTestCase):
+    def test_empty_path(self):
+        dx_p = DXPath('dx://')
+        with pytest.raises(NotImplementedError):
```
jtratner: `NotImplementedError` doesn't make sense. Perhaps this should always return True? Are there other cases where `exists()` causes an error?

stor/tests/test_dx.py
```diff
@@ -813,6 +893,11 @@ def test_project_does_not_exist(self):
 
 
 class TestGlob(DXTestCase):
+    def test_empty_path(self):
+        dx_p = DXPath('dx://')
+        with pytest.raises(ValueError, match='not supported'):
+            dx_p.glob('*')
```
jtratner: wouldn't the glob apply to project names generated?

stor/dx.py
anujkumar93: Nope no reason. Fixed

stor/dx.py
anujkumar93: This will only list those public projects where the user was specifically given permission. So, if Myriad_Demo was a public project, it would show that in the list (because we were specifically invited into it) and won't show all other public projects.

stor/dx.py
anujkumar93: Added

stor/tests/cassettes_py2/TestCopyTree/test_empty_path.yaml
anujkumar93: This call is a result of 'dest.canonical_project` [here](https://github.com/counsyl/stor/blob/master/stor/dx.py#L614). We call `_clonetree` and `_movetree` in `TestCopyTree.test_empty_path`. If it was reversed to `self.canonical_project == dest.canonical_project`, we'd not be making that dx call.

stor/dx.py
anujkumar93: Changed.

stor/dx.py
anujkumar93: Changed

stor/dx.py
anujkumar93: Changed

stor/dx.py
anujkumar93: Because it seems like overkill. We introduced empty path to just be able to see projects a user has access to. list() would involve invoking list on each of those projects (which is ~50 times for some users). List might take too long to return even. Can do `stor ls dx:// | xargs -L1 stor list` if really need to

stor/tests/test_dx_path_compat.py
anujkumar93: Added

stor/tests/test_dx.py
anujkumar93: Fixed.

stor/tests/test_dx.py
anujkumar93: I didn't implement specifically. Just canonical_project is not implemented, and this hits that. Just included in test to make it explicit

stor/tests/test_dx.py
anujkumar93: Not afaik. I don't know if it makes sense to call exists on `dx://`. Sounds like something the user is doing by mistake and should be notified of. Can fix this if you feel strongly about it

stor/tests/test_dx.py
anujkumar93: Glob is implemented internally as walkfiles. Meaning it only finds the files within the path. So ideally we should be finding all the files a user has access to across projects and returning the ones that match the glob pattern. Implementing it this way may introduce some confusion.

stor/dx.py
anujkumar93: Added

stor/tests/test_dx.py
jtratner: can imagine it with things like:

```real_paths = [f.exists() for f in paths]``` or something

anujkumar93: Gotcha. The code snipped above is outdated. This is currently fixed.

PR #115, opened 2019-06-03
Fix required version of six package.

pkaleta:
@jtratner cc @anujkumar93 

`six.raise_from`, which we use a lot was added in `six-1.9.0`. We need to require at least this version in requirements.txt.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1,6 +1,10 @@
 Release Notes
 =============
 
+v3.0.4
+------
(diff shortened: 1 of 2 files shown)


pkaleta:
correct

pkaleta:
However I looked at all the MWH repositories that pin the version of `six` to sth <= 1.8.0 in their requirements files and install stor as well. There was only wetlab-ops that did that.

PR #117, opened 2019-07-17
Add content_type property for OBS paths

jtratner:
@anujkumar93  @pkaleta - adds `content_type` property for all OBS paths.

Kinda annoying to figure out what content-type header should be set to
consistently - adding method helps with that.

Sem-Ver: feature

### DIFF
--- a/stor/dx.py
+++ b/stor/dx.py
@@ -1087,6 +1087,11 @@ class DXPath(OBSPath):
         return dxpy.DXFile(dxid=self.canonical_resource,
                            project=self.canonical_project).describe()
 
+    @property
+    def content_type(self):
+        """Get content type for DXObject. Returns empty string if not present or is project/"""
+        return self.stat().get('media') or ''
+
 
 class DXVirtualPath(DXPath):
     """Class Handler for DXPath of form 'dx://MyProject:/a/b/c' or 'dx://project-{uuid}:/b/c'"""
--- a/stor/s3.py
+++ b/stor/s3.py
@@ -468,6 +468,11 @@ class S3Path(OBSPath):
         }
         return response
 
+    @property
+    def content_type(self):
+        """Get content type for path (using ContentType field from Boto) or empty string."""
+        return self.stat().get('ContentType', '')
+
     def read_object(self):
         """Read an individual object from OBS.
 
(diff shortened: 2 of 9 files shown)


stor/dx.py
```diff
@@ -1087,6 +1087,11 @@ def stat(self):
         return dxpy.DXFile(dxid=self.canonical_resource,
                            project=self.canonical_project).describe()
 
+    @property
+    def content_type(self):
+        """Get content type for DXObject. Returns empty string if not present or is project/"""
+        return self.stat().get('media') or ''
```
anujkumar93: Note that `DXPath.stat()` raises error for a folder path. Do we also want that behavior for `content_type`?

stor/tests/test_dx.py
```diff
@@ -1830,3 +1830,30 @@ def mock_header(r, security_context):
...
+            u'modified': 1563329955907,
+            u'name': u'somefile.txt',
+            u'project': u'project-123456823409234902309423',
+            u'size': 6,
+            u'sponsored': False,
+            u'state': u'closed',
+            u'tags': [],
+            u'types': []
+        }
+        self.assertEqual(DXPath('dx://P:/C/T').content_type, 'text/csv')
```
anujkumar93:
```suggestion
        self.assertEqual(DXPath('dx://P:/C/T').content_type, 'text/plain')
```

stor/s3.py
```diff
@@ -468,6 +468,11 @@ def stat(self):
         }
         return response
 
+    @property
+    def content_type(self):
+        """Get content type for path (using ContentType field from Boto) or empty string."""
+        return self.stat().get('ContentType', '') or ''
```
pkaleta: nit: is `or ''` needed? doesn't `.get('ContentType', '')` already handle that?

counsyl-opensource:
I think I could remove the (, ‘’), but basically this was to make sure it’s not None

________________________________

counsyl-opensource:
We have a couple services that attempt to proxy on top of OBS objects - so useful to be able to pull through headers (granted if we were using underlying library this would be trivial because you could copy headers over). Just figured rather than put the code into the services directly, we’d put it into the library itself.

________________________________

stor/dx.py
jtratner: yeah that seems reasonable

stor/s3.py
jtratner: sure changed it to how you had it

PR #121, opened 2019-08-03
added readable, writable, and seekable methods to OBSFile to fully implement the io.IOBase interface

sampearl:
- added `readable`, `writable`, and `seekable` methods to `OBSFile`
- tests that OBS objects opened for reading and writing:
  - have the readable/writable/seekable methods
  - returns True for seekable
  - returns True for `readable`, False for `writable` when opened for reading
  - returns False for `readable`, True for `writable` when opened for writing

### DIFF
--- a/stor/tests/shared_obs.py
+++ b/stor/tests/shared_obs.py
@@ -211,17 +211,11 @@ class SharedOBSFileCases(object):
     def test_readable_writable_seekable(self):
         pth = self.normal_path
         read_obj = pth.open(mode='r')
-        self.assertTrue(hasattr(read_obj, 'readable'))
-        self.assertTrue(hasattr(read_obj, 'writable'))
-        self.assertTrue(hasattr(read_obj, 'seekable'))
         self.assertTrue(read_obj.readable())
         self.assertFalse(read_obj.writable())
         self.assertTrue(read_obj.seekable())
 
         write_obj = stor.open(stor.join(self.drive, 'B/C/obj'), 'w')
-        self.assertTrue(hasattr(write_obj, 'readable'))
-        self.assertTrue(hasattr(write_obj, 'writable'))
-        self.assertTrue(hasattr(write_obj, 'seekable'))
         self.assertFalse(write_obj.readable())
         self.assertTrue(write_obj.writable())
         self.assertTrue(write_obj.seekable())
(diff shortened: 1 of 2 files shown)


sampearl:
@tophers42 @kyleabeauchamp @jtratner Could you all give this a look?

sampearl:
Specifically I wasn't sure whether I put my test in the proper place (`shared_obs`), or that it would be properly parameterized/used by each type of OBS file system test using `self.normal_path` and `self.drive`.

kyleabeauchamp:
I'll need another look from @jtratner  et al then

jtratner:
Looks great - thanks!

I have one small request (that I can also do before merging if you’re okay
with it): let’s get rid of the assertTrue(hasattr checks - the next couple
lines cover the same thing ;)

Jeff

stor/tests/shared_obs.py
```diff
@@ -207,3 +207,21 @@ def test_seek_behavior_and_writes(self, mock_write_object):
             obj.seek(0)
         mock_write_object.assert_called_with(self.normal_path, b'happy-go_lucky')
         self.assertEquals(len(mock_write_object.call_args_list), 1)
+
+    def test_readable_writable_seekable(self):
+        pth = self.normal_path
+        read_obj = pth.open(mode='r')
+        self.assertTrue(hasattr(read_obj, 'readable'))
```
tophers42: Nit: Split each on into a separate test.

jtratner: I disagree with this nit - it's great to have 'em all together, I think really nbd

tophers42: Unit tests should test a unit. These functions are separate units, therefore separate tests. of course they can all be next to each other.

sampearl:
@jtratner re: `hasattr` - sure, if it doesn't have those methods it'll fail on the method invocation in a reasonable/useful way?

stor/tests/shared_obs.py
sampearl: I'm fine with it either way.

jtratner:
@sampearl  - yep - see below

```
In [1]: import stor
stor
In [2]: stor.Path("whatever")
Out[2]: PosixPath('whatever')

In [3]: _.reaaadable()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-3b3cec0e2a4b> in <module>
----> 1 _.reaaadable()

AttributeError: 'PosixPath' object has no attribute 'reaaadable'
```

jtratner:
Thanks for the PR @sampearl !

sampearl:
👍 !

PR #125, opened 2019-09-26
Fix logic around calling wait_on_close

jtratner:
1. Only call it when we've written data (i.e., there's a buffer) - fixes
   issues where destructor was throwing errors. (came up when watching travis logs)
2. Only call it for WRITE_MODES (this should potential issue where
   holding a DX file object in read mode would cause an extra API
   request).

I'd like to double check this, but my belief on reading this is that we were previously making an additional API request to DNAnexus every time we attempted to read a DXFile object (rather than only doing it on writes).

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -12,6 +12,13 @@ API Additions
 * Added ``content_type`` property for OBS paths representing consistent
   Content-Type header that would be set on download.
 
+Bug Fixes
+^^^^^^^^^
+
+* Only call ``OBSFile._wait_on_close()`` when we actually wrote data AND we
+  were in write mode. Also fixes issues where malformed ``OBSFile`` objects
+  would throw an exception in ``__del__`` method.
+
 v3.0.5
 ------
 * Changed ``dxpy3`` requirement to ``dxpy`` to reflect new Python3 compatible ``dxpy`` module.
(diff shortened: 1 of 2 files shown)


jtratner:
Yay - no more of these errors!

Previous:

```
test_append_mode_fail (stor.tests.test_dx.TestOpen) ... Exception ignored in: <bound method OBSFile.__del__ of <stor.obs.OBSFile object at 0x7f0ea4560080>>
  File "/home/travis/build/counsyl/stor/stor/obs.py", line 408, in __del__
    self.close()
  File "/home/travis/build/counsyl/stor/stor/obs.py", line 464, in close
    self._wait_on_close()
  File "/home/travis/build/counsyl/stor/stor/obs.py", line 467, in _wait_on_close
    if isinstance(self._path, stor.dx.DXPath):
AttributeError: 'OBSFile' object has no attribute '_path'
```

New:

```
test_append_mode_fail (stor.tests.test_dx.TestOpen) ... ok
```

woo!

jtratner:
@anujkumar93 - if you end up regenerating cassettes would be interesting to see if we have one fewer API request with this (and whether worth writing a test case for that portion of the behavior).

Otherwise relatively hard to test "did not error out in __del__" :-/

jtratner:
maybe a test with a mock that shows read mode no longer calls `wait_on_close` ever!

PR #126, opened 2019-09-26
For discussion: Do not block when garbage collecting OBSFile's

jtratner:
Really only makes sense in ``__exit__()`` (i.e., when used in ``with``
statement).

Otherwise I think you can hang main thread in a `gc` block by accident.

Thoughts on this @anujkumar93 @pkaleta ?


jtratner:
Failure unrelated - merging.

jtratner:
NOTE to self, gotta move the doc note to next version

PR #128, opened 2020-04-04
Fix release notes

jtratner:
@pkaleta - just small tweak to get release notes associated with proper version


jtratner:
Yep it's a pending release

PR #129, opened 2020-04-04
Add a `stor completions` command to generate tab completion text

jtratner:
@pkaleta - in prep for switching to poetry as build tool (which makes it easier to do github actions anyways)

`poetry` does not support bash scripts in its `scripts` section. So
instead we add a `stor completions` command to dump to stdout (which
accomplishes essentially same purpose as previous)


jtratner:
@pkaleta - this is actually included in the midst of #130  - just split it out because it's actually a material change

PR #130, opened 2020-04-05
Python 3 + Poetry + Github Actions testing

jtratner:
@pkaleta cc @anujkumar93 - I noticed Travis CI was no longer polling repo, so I took. a bit of time to upgrade stor testing suite to use github actions.

### Changes

* **Switch to Github Actions**  - see workflow file - faster than travis / more integrated / has caching
* **Switch to poetry for build tool** - While trying to use github actions, I hit issues with virtualenv and friends, so I figured "Why not just upgrade to poetry because it's easier?"  So... I did!
* **Drop Python 2 support** Finally, I realized that I was changing a bunch of things anyways, so I may as well drop Python 2 support (now that it's EOL anyways) so we can focus on developing with nice type annotations and such.
* **Drop support for python 3.5** - We want to use f-strings.
* **Fix hidden lint issues** Lint issues are because (as usual) our version of flake8 was so out of date that it was just silently ignoring everything.

### Future

* **blackify repo** - tired of dealing with dumb lint errors
* **mypy type annotations** - similarly, we should be doing types and such
* **enforce import sorting** - some way, preferably with automatic reordering.


jtratner:
NOTE: one issue (currently) is that we will have to manually tag releases (similar to `samwell`).  Ideally we'll make some kind of default action that auto-packages release notes and increments version BUT... for now this is what we got.

jtratner:
Nvm - ditched python3.5 cuz I'd already used f strings

setup.cfg
```diff
@@ -11,44 +11,23 @@ classifier =
     Topic :: Internet :: WWW/HTTP :: Dynamic Content
     Intended Audience :: Developers
     Programming Language :: Python
-    Programming Language :: Python :: 2
-    Programming Language :: Python :: 2.7
     Programming Language :: Python :: 3
     Programming Language :: Python :: 3.5
```
pkaleta, line 15: Remove?

setup.cfg
```diff
@@ -11,44 +11,23 @@ classifier =
...
 [flake8]
-max-complexity=10
+max-complexity=13
 max-line-length=99
 exclude = stor/third_party/*
-
-[pbr]
-skip_authors = true
-skip_changelog = true
+ignore=E226,W504
```
pkaleta, line 30: were these causing many issues? I personally like it

stor/s3.py
```diff
@@ -601,7 +594,7 @@ def download(self, dest, condition=None, use_manifest=False, **kwargs):
                     except StopIteration:
                         break
                 pool.close()
-            except:
+            except BaseException:
```
pkaleta, line 597: is there a difference between this and previous version?

https://stackoverflow.com/a/7161517

stor/tests/test_posix_path_compat.py
```diff
@@ -179,7 +179,7 @@ def test_listing(self):
                 for f in files:
                     try:
                         f.remove()
-                    except:
+                    except Exception:
```
pkaleta, line 182: This will no longer catch special Python exceptions https://stackoverflow.com/a/7161517. Is this expected?

pkaleta:
I've noticed there are some annotations failures, have you seen this?

Screen Shot 2020-04-06 at 2 32 57 PM

stor/s3.py
jtratner: No, but linter complains otherwise. Also I appreciate that it indicates to reader we want to catch KeyboardInterrupt and friends

stor/tests/test_posix_path_compat.py
jtratner: Yes in this case I don’t think we meant to catch KeyboardInterrupt or SystemExit :)

setup.cfg
jtratner: Like I said, linter had been swallowed for a while. You are totally welcome if you want to remove this and fix lint issues that arise :)

jtratner:
> I've noticed there are some annotations failures, have you seen this?
> 
> Screen Shot 2020-04-06 at 2 32 57 PM

This is because of an expected traceback that gets dumped in a test. Would be good to stop emitting it. Def good fodder for ticket

setup.cfg
jtratner: Have follow up PR to remove this file completely

jtratner:
One more PR incoming with some additional tweaks to get poetry fully working :j

PR #133, opened 2020-04-06
Error earlier when open() called on DXProject to fix uncaught exception in `__del__ `

jtratner:
Raise an exception directly in `stor.open` if we have a project path.

Previously, we were raising an exception at `write_object`;
however, if you actually attempted to write data to the path, this would
result in an exception in the `__del__` method as `stor` tried to flush
data.

This also helps reduce noise from the test suite (i.e., those test annotations)

Documented changes:

* Eliminate exception in ``__del__`` method when calling ``stor.open`` on a DNAnexus project path.
* ``stor.open()`` now throws an error earlier when incorrectly calling ``open()`` on a DNAnexus project path.

### DIFF
--- a/stor/dx.py
+++ b/stor/dx.py
@@ -848,6 +848,34 @@ class DXPath(OBSPath):
             suo = OBSUploadObject(fp.name, object_name='/' + self.resource)
             return self.upload([suo], **kwargs)
 
+    def open(self, mode='r', encoding=None):
+        """
+        Opens a OBSFile that can be read or written to and is uploaded to
+        the remote service.
+
+        For examples of reading and writing opened objects, view
+        OBSFile.
+
+        Args:
+            mode (str): The mode of object IO. Currently supports reading
+                ("r" or "rb") and writing ("w", "wb")
+            encoding (str): text encoding to use. Defaults to
+                ``locale.getpreferredencoding(False)``
+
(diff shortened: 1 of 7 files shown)


stor/dx.py
```diff
@@ -848,6 +848,34 @@ def write_object(self, content, **kwargs):
...
+    def open(self, mode='r', encoding=None):
+        """
+        Opens a OBSFile that can be read or written to and is uploaded to
+        the remote service.
+
+        For examples of reading and writing opened objects, view
+        OBSFile.
+
+        Args:
+            mode (str): The mode of object IO. Currently supports reading
```
pkaleta, line 860: nit: since this is now using Python 3, can you move type annotations to the method def?

stor/dx.py
```diff
@@ -848,6 +848,34 @@ def write_object(self, content, **kwargs):
...
+                ("r" or "rb") and writing ("w", "wb")
+            encoding (str): text encoding to use. Defaults to
+                ``locale.getpreferredencoding(False)``
+
+        Returns:
+            OBSFile: The file object for Swift/S3/DX.
+
+        Raises:
+            ValueError: if attempting to write to project
```
pkaleta, line 869: nit: if attempting to call ``open()`` on a project

PR #134, opened 2020-04-08
Update temp_url helptext when filename=''

jtratner:
@pkaleta - please take a look - updated docstring with your suggestion

### DIFF
--- a/stor/dx.py
+++ b/stor/dx.py
@@ -203,7 +203,7 @@ class DXPath(OBSPath):
                 URL will be valid (only for temp URL generation)
             filename (str, optional): A urlencoded filename to use for
                 attachment, otherwise defaults to object name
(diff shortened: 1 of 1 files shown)


stor/dx.py
```diff
@@ -203,7 +203,7 @@ def temp_url(self, lifetime=300, filename=None):
                 URL will be valid (only for temp URL generation)
             filename (str, optional): A urlencoded filename to use for
                 attachment, otherwise defaults to object name
-                (to bypass, set to empty string).
+                (to use no filename at all, use ``filename=''``).
```
jtratner, line 206:
```suggestion
                (to use no filename at all, use ``filename=''``)
```

jtratner:
@pkaleta - looks like the `pull_request` event _does_ fire for every commit to a PR :)

PR #136, opened 2020-04-26
Add trailing slash to file_proxy_url if absent.

pkaleta:
@jtratner 

Right now, [this example from the docs](https://counsyl.github.io/stor/dx.html#stor.dx.DXPath.temp_url) doesn't work, because proxy URL doesn't have a trailing slash and therefore the `gateway` part is stripped off of the generated URL:



```
In [3]: with stor.settings.use({'dx': {'file_proxy_url': 'https://my-dnax-proxy.example.com/gateway'}}): 
   ...:     print(stor.Path('dx://proj:/folder/mypath.csv').temp_url()) 
   ...:                          
https://my-dnax-proxy.example.com/proj/folder/mypath.csv
```

It does work however, if you add a trailing slash to the proxy URL:

```
In [2]: with stor.settings.use({'dx': {'file_proxy_url': 'https://my-dnax-proxy.example.com/gateway/'}}): 
   ...:     print(stor.Path('dx://proj:/folder/mypath.csv').temp_url()) 
   ...:                                                                                                                                                                                                                                
https://my-dnax-proxy.example.com/gateway/proj/folder/mypath.csv
```

This PR adds a trailing slash to the `file_proxy_url` if it's missing. I've also added test that replicates the issue and proves that this fix works.

### DIFF
--- a/stor/tests/test_dx.py
+++ b/stor/tests/test_dx.py
@@ -1005,6 +1005,8 @@ class TestTempUrl(DXTestCase):
                 with pytest.raises(ValueError, match='filename MUST match object name'):
                     dx_p.temp_url(filename='another_name.txt')
 
+            assert proxy_path in dx_p.temp_url()
+
     def test_on_file_with_proxy_url(self):
         self.setup_temporary_project()
         self.setup_file('/temp_file.txt')
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1,12 +1,17 @@
 Release Notes
 =============
 
+v4.0.1
+------
+
+* Add trailing slash to file_proxy_url if absent, to fix a bug with generating temp URLs.
+
 v4.0 - Python 3 + Poetry!
 -------------------------
 
 Only Python 3.6+ is supported - Python 2.7 and Python 3.5 support is dropped!
 
-Note: the major version bump is because we've dropped 2.7 support, aside from that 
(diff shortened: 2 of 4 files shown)


jtratner:
I wonder if we should throw exception instead? Is there case where this would be wrong?

pkaleta:
>Is there case where this would be wrong?

Hm, I can't imagine a case where you'd specify `https://my-gateway.foo.com/gateway`, but didn't want the `gateway` part to be included in the generated URLs.

If we decide to fail loudly, we probably should do this when configuring stor (via `stor.settings.update`) and when reading from `DX_FILE_PROXY_URL` env var, rather than when `temp_url()` is actually called.

WDYT?

stor/tests/test_dx.py
```diff
@@ -1005,6 +1005,8 @@ def _test_proxy_url(self, proxy_path):
                 with pytest.raises(ValueError, match='filename MUST match object name'):
                     dx_p.temp_url(filename='another_name.txt')
 
+            assert proxy_path in dx_p.temp_url()
```
jtratner, line 1008: Needs something where it doesn’t have trailing slash on it?

pkaleta: it already has such example:

```
self._test_proxy_url("https://myproxy.example.com/dnax-gateway")
```

down below in `test_on_file_with_proxy_url`

PR #138, opened 2020-08-15
add read mode for dxpy

ying-w:
Now that [dxpy](https://github.com/dnanexus/dx-toolkit/issues/426) supports non-utf8 read modes, pass 'rb' through

`make lint` doesn't work tho:

```
Warning, treated as error:
failed to reach any of the inventories with the following issues:
intersphinx inventory 'http://docs.python.org/3/objects.inv' not fetchable due to <class 'requests.exceptions.ConnectionError'>: HTTPConnectionPool(host='docs.python.org', port=80): Max retries exceeded with url: /3/objects.inv (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x107478d50>: Failed to establish a new connection: [Errno 60] Operation timed out'))
```

### DIFF
--- a/stor/obs.py
+++ b/stor/obs.py
@@ -128,7 +128,7 @@ class OBSPath(Path):
         normed = posixpath.normpath('/' + str(self)[len(self.drive):])[1:]
         return self.path_class(self.drive + normed)
 
-    def read_object(self):
+    def read_object(self, mode='r'):
         """Reads an individual object from OBS.
 
         Returns:
@@ -413,7 +413,7 @@ class OBSFile(object):
         if self.mode == 'r':
             buf = self.stream_cls(self._path.read_object().decode(self.encoding))
         elif self.mode == 'rb':
-            buf = self.stream_cls(self._path.read_object())
+            buf = self.stream_cls(self._path.read_object(mode='rb'))
         elif self.mode in ('w', 'wb'):
             buf = self.stream_cls()
         else:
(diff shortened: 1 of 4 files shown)


stor/obs.py
```diff
@@ -128,7 +128,7 @@ def normpath(self):
         normed = posixpath.normpath('/' + str(self)[len(self.drive):])[1:]
         return self.path_class(self.drive + normed)
 
-    def read_object(self):
+    def read_object(self, mode='r'):
```
anujkumar93, line 131: If the intention is to use `with stor.Path('dx://path/to/file).open(mode='rb')` to access dx files, this is not really necessary because mode is already set in the `self` object by this point. You can remove it from `dx.py`, `s3.py` and `swift.py` as well. You'd just do in `dx.py`:
```
def read_object(self):
    ....
    file_handler = dxpy.DXFile(dxid=self.canonical_resource, project=self.canonical_project, mode=self.mode)
    ....
    if self.mode == 'r':
        ....
```

If there was some other reasoning behind this, you can disregard my comment.

ying-w: Thanks for the feedback. I just realized that `OBSFile` defined `mode` i was looking for it in `OBSPath` instead and not finding it.

I'll make the change but I'm not really sure about when when to use `open()` vs `read_object()`

jtratner:
I think that the correct answer is to just not decode? Probably just need to review this again - good point...

PR #140, opened 2020-09-18
fix DeprecationWarning

ying-w:


jtratner (approved):
lgtm - this seems safe given that in python 3 it's the same and python 2 is mostly deprecated

PR #142, opened 2021-04-13
Support mixed canonical paths

jtratner:
Currently when you try to use a path like `dx://someproj:file-123`, you get an error about data not found:

```
NotFoundError: No data object was found for the given path (dx://myproj:file-G0GYf0Q0BqQ4kbBfJvGgFG4K) on DNAnexus
```

This just fixes it to check if valid dxid and then returns it.


pkaleta (approved):
lgtm

PR #144, opened 2023-06-16
Update stor upload, download to work on AWS lambda for python3.8 runtime

scdunatun:
### Description

AWS lambda cannot use `multiprocessing.Pool` and `multiprocessing.Queue` as of python3.8, due to the use of `SemLock` (see [this blog post from AWS](https://aws.amazon.com/blogs/compute/parallel-processing-in-python-with-aws-lambda/)). In order to use `stor` on AWS lambda with a python3.8 runtime, we will replace use of `multiprocessing.ThreadPool` with `concurrent.futures.ThreadPoolExecutor`, which does not use SemLock.

### Changes
 - Updated `ThreadPool` to `ThreadPoolExecutor` in `stor/s3.py` for both `upload` and `download`.
 - Updated related unit tests for upload and download.
   - Added `MockFuture` and `MockExecutor` for mocking `ThreadPoolExecutor`.
   - Added `tearDown()` method to `test.py::S3TestCase` to prevent mock teardown/cleanup bleed. 
 - Updated `test_upload_other_error` to appropriately raise `FailedUploadError` with `ThreadPoolExecutor` changes.
 - Updated `test_download_other_error` to appropriately raise `FailedDownloadError` with `ThreadPoolExecutor` changes.

### Testing
 - Successful execution of updated tests:
    - `test_upload_object_threads`
    - `test_upload_other_error`
    - `test_download_object_threads`
    - `test_download_other_error`
- Successful execution of CI testing on all three python versions. 

### Todo
- [x] Update version prior to merge.
- [x] Any necessary docstring updates for s3.py `upload` and `download`.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -7,10 +7,10 @@ env:
 jobs:
   test:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8]
 
     steps:
     - uses: actions/checkout@v2
--- a/stor/s3.py
+++ b/stor/s3.py
@@ -3,7 +3,7 @@ An experimental implementation of S3 in stor
 """
 from functools import partial
 import logging
-from multiprocessing.pool import ThreadPool
+from concurrent.futures import as_completed, ThreadPoolExecutor
 import os
 import tempfile
 import threading
@@ -580,25 +580,18 @@ class S3Path(OBSPath):
 
         downloaded = {'completed': [], 'failed': []}
         with S3DownloadLogger(len(files_to_download)) as dl:
-            pool = ThreadPool(options['object_threads'])
-            try:
-                result_iter = pool.imap_unordered(download_w_config, files_to_download)
-                while True:
-                    try:
-                        result = result_iter.next(0xFFFF)
-                        if result['success']:
-                            dl.add_result(result)
-                            downloaded['completed'].append(result)
-                        else:
-                            downloaded['failed'].append(result)
-                    except StopIteration:
-                        break
-                pool.close()
-            except BaseException:
-                pool.terminate()
-                raise
-            finally:
-                pool.join()
+            with ThreadPoolExecutor(max_workers=options.get("object_threads")) as executor:
+                futures = {
+                    executor.submit(download_w_config, file_to_download): file_to_download
+                    for file_to_download in files_to_download
+                }
+                for fut in as_completed(futures):
+                    result = fut.result()
+                    if result["success"]:
+                        dl.add_result(result)
+                        downloaded['completed'].append(result)
+                    else:
+                        downloaded['failed'].append(result)
 
         if downloaded['failed']:
             raise exceptions.FailedDownloadError('an error occurred while downloading', downloaded)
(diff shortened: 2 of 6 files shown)


stor/tests/test_s3.py
```diff
@@ -1085,19 +1108,21 @@ def test_upload_w_use_manifest_single_file(self, mock_getsize, mock_files):
...
             for i in range(20)
         }
         mock_getsize.return_value = 20
-        mock_pool.return_value.imap_unordered.return_value.next.side_effect = StopIteration
 
         s3_p = S3Path('s3://bucket')
         with settings.use({'s3:upload': {'object_threads': 20}}):
             s3_p.upload(['test'])
-        mock_pool.assert_called_once_with(20)
+        mock_pool.assert_called_once_with(max_workers=20)
```
Ben-Habermeyer, line 1125: this test was previously pretty simple just asserting that the mock pool was called with the correct args -> can we do something like the following instead?

```
# either in decorator or context manager -> mock object and spec
mock_pool = mock.patch("stor.s3.ThreadPoolExecutor", autospec=True)
mock_executor = mock.patch("stor.s3. as_completed", autospec=True)

# return a function which takes a callable and val and returns a val
# or can just hard code .return_value = 5 or something
mock_pool.return_value.submit.side_effect = lambda fun, val: val 

# return a list of the length of input - can be anything with .result() function
mock_as_completed.side_effect = lambda futures: [MockFuture for future in futures]


# and assert the mocks were called with right args etc.
```

.github/workflows/pythonpackage.yml
```diff
@@ -7,7 +7,7 @@ env:
 jobs:
   test:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
Ben-Habermeyer, line 10: I'm guessing old actions weren't saved to see what the runtime was before

Ben-Habermeyer:
we will also need the PBR semver bump, the release notes and version bump in https://github.com/counsyl/stor/pull/141/files -> with these changes I think `4.1.0`

.github/workflows/pythonpackage.yml
scdunatun: Too old; the logs are all cleaned. Based on a quick search, 20.04 is the last one that supports py36, arch x64 though, so that's why I put it here.

stor/tests/test_s3.py
scdunatun: Based on how the submit mocking was going, I left it in as calling the mock class. Added additional asserts for as_completed being called correctly in [834443b](https://github.com/counsyl/stor/pull/144/commits/834443bc0df6f9ce9d0718372cbd81fc865e7259).

scdunatun:
> we will also need the PBR semver bump, the release notes and version bump in https://github.com/counsyl/stor/pull/141/files -> with these changes I think `4.1.0`

Completed in 41df381; pretty sure the versioning is controlled directly in pyproject.toml, so I bumped version with `poetry version minor` instead of doing a PBR versioning commit.

stor/test.py
```diff
@@ -345,3 +350,38 @@ def teardown_posix_files(self):
...
+    def __init__(self):
+        self._shutdown = False
+        self._shutdownLock = Lock()
+
+    def submit(self, fn, *args, **kwargs):
+        obj = args[0]
+        if isinstance(obj, dict):
+            dest = obj.get("dest")
+            source = obj.get("source")
+        else:
```
scdunatun, line 379: I could probably be more specific that this is the upload object...

Ben-Habermeyer: yea I think it would be helpful to add a docstring for these functions describing why you added them i.e. mocking executor.submit or what they do

stor/s3.py
```diff
@@ -580,25 +580,18 @@ def download(self, dest, condition=None, use_manifest=False, **kwargs):
...
-                raise
-            finally:
-                pool.join()
+            with ThreadPoolExecutor(max_workers=options.get("object_threads")) as executor:
+                futures = {
+                    executor.submit(download_w_config, file_to_download): file_to_download
+                    for file_to_download in files_to_download
+                }
+                for fut in as_completed(futures):
+                    result = fut.result()
```
Ben-Habermeyer, line 589: any way to catch an exception here and return the file being downloaded which is problematic in the stack trace?
```suggestion
try:
   result = fut.result()
except Exception as e:
    file = future.something(). # not sure how to get file_to_download
    raise SomeGoodErrorMEssageWithFile
```

stor/test.py
scdunatun: Definitely wanted to add more documentation, which I shall do soon!

scdunatun:
> I think the tests lgtm - anything else to do here before ready for review?

I'll address your earlier comments (I especially like the one about catching an exception and seeing if we can return the bad file(s)) and then open this for official review.

stor/s3.py
scdunatun: Completed for download in 04cb7b8
Then for upload (with some download test updates) in bcb680d

stor/test.py
scdunatun: Added some additional documentation in 2849a72.

Happy to add more if you feel it is not clear enough!

stor/s3.py
```diff
@@ -585,16 +585,26 @@ def download(self, dest, condition=None, use_manifest=False, **kwargs):
                     executor.submit(download_w_config, file_to_download): file_to_download
                     for file_to_download in files_to_download
                 }
-                for fut in as_completed(futures):
-                    result = fut.result()
+                for fut in as_completed(futures.keys()):
+                    source, dest = futures[fut]
```
Ben-Habermeyer, line 589: is this a tuple? or is it a dict `{'source': file, 'dest': dest}`

Ben-Habermeyer (approved):
one question about the error raising but otherwise LGTM -> would be nice to test this one within lambda if we can install the dist and copy it in. I tagged you on a Teams thread

stor/s3.py
scdunatun: It should be a dict. Looks like I did not add that update.

Basically we won't have this line and the exception should call `f'{futures[fut]["source"]}: {e}'` directly instead.

scdunatun: Fixed in [b496d1a](https://github.com/counsyl/stor/pull/144/commits/b496d1a326f921591a00cacfa79a4125c8224286)

Ben-Habermeyer (approved):
update LGTM

PR #149, opened 2024-04-19
Fix tests by removing deprecated python 3.6 from test matrix

jtratner:
As it says in description, replace python 3.6 with 3.9, now that 3.6 is no longer available as a simple built-in environment on github actions.

Sadly 3.12 seemed to have an issue with installing here, so saving that for a future improvement :(


Ben-Habermeyer (approved):
LGTM thanks for the update!

PR #151, opened 2025-12-30
[BIOINFSW-6193] Replace pkg_resources use with importlib metadata

scdunatun:
## Description
`pkg_resources` is being deprecated and all applications that install stor are unable to upgrade the setuptools version due to its use of `pkg_resources`.  Instead of `pkg_resources`, we should use `importlib`'s `metadata`.

### Changes
- Updated use of `pkg_resources` to `importlib`'s `metadata` in `__init__.py`. Fall back import and use for older version of python.
- Updated CI/CD to run on `ubuntu-22.04` runners with updated action version tags.

### Testing
- CI/CD successful for python 3.7 and 3.8.

### DIFF
--- a/stor/__init__.py
+++ b/stor/__init__.py
@@ -20,7 +20,15 @@ functions.
 
 See `stor.swift` for more information on Swift-specific functionality.
 """
-from importlib import metadata
+try:
+    from importlib import metadata
+except ImportError:  # Python < 3.8
+    try:
+        import importlib_metadata as metadata
+    except ImportError:
+        # Fall back to pkg_resources for Python < 3.8 if importlib_metadata not available
+        import pkg_resources
+        metadata = None
 
 from stor.utils import copy
 from stor.utils import copytree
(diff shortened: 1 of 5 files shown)


stor/__init__.py
```diff
@@ -20,7 +20,15 @@
...
-import pkg_resources
+try:
+    from importlib import metadata
+except ImportError:  # Python < 3.8
+    try:
+        import importlib_metadata as metadata
+    except ImportError:
+        # Fall back to pkg_resources for Python < 3.8 if importlib_metadata not available
+        import pkg_resources
+        metadata = None
```
Ben-Habermeyer: @scdunatun I didn't want to add an `importlib_metadata` dependency to `stor`.. so this is essentially backwards compatibility.

For python >=3.8 `importlib` will be installed and use that

For python <3.8 try using `importlib_metadata` (backport) if it is installed, otherwise use the fallback `pkg_resources`

Once we deprecate <3.10 we can remove all of this i favor of importlib

Ben-Habermeyer (approved):
LGTM. Should be backwards compatible for poor souls on 3.6/7 and fix deprecation warning for 3.8+.

We can make a story to release a new major version deprecating these old versions and modernizing the ci

